PR #19, opened 2015-02-12
Adds tags as an optional package field.

ConnorDoyle:


repo/meta/schema/package-schema.json
```diff
@@ -65,6 +65,13 @@
           "version"
         ]
       }
+    },
+    "tags": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "pattern": "[a-z]+"
```
jsancio: why the restriction of only lower case letters? For example users want to tag something as `python3`.

ConnorDoyle: Good call, will open this up to allow numbers as well.

ConnorDoyle: To be even less restrictive, we'll allow everything except whitespace here.

ConnorDoyle: Done, just needs to be a nonempty string without whitespace.

PR #23, opened 2015-02-17
Added labels to the marathon schema.

ConnorDoyle:

PR #24, opened 2015-02-18
Fix the Marathon schema for labels

jsancio:
To alloy any property that maps to a string we need to use
`additionalProperties`. See
http://spacetelescope.github.io/understanding-json-schema/reference/object.html#properties


ConnorDoyle:
Yep, that's what I meant to write in #23.  Nice catch!

PR #26, opened 2015-03-02
Added package for Marathon.

ConnorDoyle:
Also disables validation of Marathon templates as JSON, since Mustache template syntax is outside of the language.


ConnorDoyle:
cc @ssk2

repo/meta/index.json
```diff
@@ -22,6 +22,18 @@
       }
     },
     {
+      "currentVersion":"0.8.1-RC1",
+      "description":"A cluster-wide init and control system for services in cgroups or Docker containers.",
```
ssk2: I'd remove the "in cgroups or Docker containers." and say something like:
`A cluster-wide init and control system for containerized services."

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
+{
+  "type": "object",
+  "properties": {
+    "marathon/libprocess-ip": {
+      "description": "Interface to bind to for communication with the Mesos master.",
+      "type": "string"
+    },
+    "mesos/master": {
+      "description": "The URL of the Mesos master. The format is a comma-delimited list of of hosts like zk://host1:port,host2:port/mesos. If using ZooKeeper, pay particular attention to the leading zk:// and trailing /mesos! If not using ZooKeeper, standard URLs like http://localhost are also acceptable.",
```
ssk2: Extra `of`

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
+{
+  "type": "object",
+  "properties": {
+    "marathon/libprocess-ip": {
```
ssk2: `libprocess` here seems like an implementation detail. How about `marathon/bind-ip`?

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
...
+      "type": "integer",
+      "minimum": 0,
+      "default": 604800
+    },
+    "marathon/framework-name": {
+      "description": "The framework name to register with Mesos.",
+      "type": "string"
+    },
+    "marathon/ha": {
+      "description": "Runs Marathon in HA mode with leader election. Allows starting an arbitrary number of other Marathons but all need to be started in HA mode. This mode requires a running ZooKeeper.",
```
ssk2: `HA` -> `high availability (HA)` (no brevity requirement here and it may not be intuitive to people what HA means)

repo/meta/index.json
ConnorDoyle: This is taken directly from the project description on GitHub.

repo/packages/M/marathon/0/config.json
ConnorDoyle: This currently has clear 1-to-1 correspondence with the project's command-line arguments.  The description should be sufficient to explain what this means.

repo/packages/M/marathon/0/config.json
ConnorDoyle: That would be quite confusing in my opinion, since it's not the interface we necessarily bind on for the API service.

repo/meta/index.json
ssk2: Okay, I disagree with that one too. Here's a PR to change that: https://github.com/mesosphere/marathon/pull/1260

repo/packages/M/marathon/0/config.json
ssk2: Fair point. How about `mesos/bind-ip` then? Whatever we do, I think it's unnecessary to include `libprocess`.

repo/packages/M/marathon/0/config.json
ssk2: Oh, I meant in the description itself, the name is fine.

repo/packages/M/marathon/0/config.json
ConnorDoyle: My idea here is that there are a bunch of docs related to the LIBPROCESS_IP environment variable, so I think it's nice to make it explicit that these are the same thing.

repo/packages/M/marathon/0/config.json
ConnorDoyle: Right, good call.

repo/packages/M/marathon/0/config.json
ssk2: Alright - that makes sense. Hopefully that'll become deprecated sometime.

repo/packages/M/marathon/0/marathon.json
```diff
@@ -0,0 +1,151 @@
+{
```
jsancio: Just a side comment that I thought about now. Technically this are not valid JSON until the mustache template engine has been applied. Should we give them a different file name? Is there are standard way of doing this for Mustache?

ConnorDoyle: Not sure, and good point. Do you have a different extension in mind?

jsancio: Had a quick look at the mustache repository and it looks like they use `*.mustache`: https://github.com/mustache/mustache/tree/master/examples

scripts/1-validate-packages.sh
```diff
@@ -23,7 +23,7 @@ validate "command.json" "$SCHEMA_DIR/command-schema.json";
 validate "config.json" "$SCHEMA_DIR/config-schema.json";
 
 # validate all marathon.json files
-validate "marathon.json" "$SCHEMA_DIR/marathon-schema.json";
+# validate "marathon.json" "$SCHEMA_DIR/marathon-schema.json";
```
jsancio: Let's not commit commented out code. Just remove it.

repo/packages/M/marathon/0/marathon.json
ConnorDoyle: Cool, created issue #27 to track this.  Will address with a subsequent PR.

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
...
+    },
+    "mesos/master": {
+      "description": "The URL of the Mesos master. The format is a comma-delimited list of of hosts like zk://host1:port,host2:port/mesos. If using ZooKeeper, pay particular attention to the leading zk:// and trailing /mesos! If not using ZooKeeper, standard URLs like http://localhost are also acceptable.",
+      "type": "string"
+    },
+    "marathon/cpus": {
+      "description": "CPU shares to allocate to each Marathon instance.",
+      "type": "number",
+      "minimum": 0.0,
+      "default": 1.0
```
jsancio: Is this true for Marathon? Can Marathon run with 0.0? How about 1.0?

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
...
+    },
+    "marathon/mem": {
+      "description": "Memory (MB) to allocate to each Marathon instance.",
+      "type": "number",
+      "minimum": 0.0,
+      "default": 512.0
```
jsancio: It is true for Marathon? Can Marathon run with 0.0 memory? How about the default 512.0?

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
...
+    },
+    "marathon/executor": {
+      "description": "Executor to use when none is specified.",
+      "type": "string"
+    },
+    "marathon/failover-timeout": {
+      "description": "The failover_timeout for Mesos in seconds. If a new Marathon instance has not re-registered with Mesos this long after a failover, Mesos will shut down all running tasks started by Marathon. Requires checkpointing to be enabled.",
+      "type": "integer",
+      "minimum": 0,
+      "default": 604800
```
jsancio: Is a minimum of 0 valid for the failover-timeout?

repo/packages/M/marathon/0/config.json
```diff
@@ -0,0 +1,147 @@
...
+      "type": "string"
+    },
+    "marathon/ssl-keystore-path": {
+      "description": "Path to the SSL keystore. SSL will be enabled if this option is supplied.",
+      "type": "string"
+    }
+  },
+  "required": [
+    "mesos/master",
+    "marathon/zk"
```
jsancio: Should all the non-required fields have a default? I assume that every field here is reference in the `marathon.json` so in that sense everything is required, no? Does required here mean that the user needs to provide a value through the json passed to `dcos package install --options=user.json`?

jsancio: I see you have {#...}...{/...} in the marathon app resource. Based on that, it looks like we require: `marathon/cpus`, `marathon/mem`, `marathon/instances`, `marathon/http-port`. It doesn't look like `meso/master` and `marathon/zk` are required.

repo/packages/M/marathon/0/config.json
ConnorDoyle: I really just mean it wants a non-negative real.  Not sure if this is the right place to guard against ridiculously low resource allocations.

repo/packages/M/marathon/0/config.json
ConnorDoyle: I really just mean it wants a non-negative real.  Not sure if this is the right place to guard against ridiculously low resource allocations.

repo/packages/M/marathon/0/config.json
ConnorDoyle: Yes.

repo/packages/M/marathon/0/config.json
jsancio: Have we thought about how these variables resolution is going to work once we implement resolving them from the `Dcos.toml` for the CLI? We already have variables under `[marathon]` and I suspect that we are going to have variables under `[mesos]`. Let's talk about this soon.

ConnorDoyle: The parameters that I didn't wrap in conditional expressions have defaults.  Could definitely drop the conditionals around the required parameters as well.

ConnorDoyle: IMO how values are marshalled in the client environment is an issue for clients to figure out, unless there are issues that will impact the spec for the package repo.

ConnorDoyle: Removed the conditionals around `mesos/master` and `marathon/zk`.

jsancio:
Looks good! :ship:

PR #34, opened 2015-04-01
renamed marathon/artifact_store to marathon/artifact-store to match the ...

swartzrock:
...general dash separation and also match marathon.json


jsancio:
Looks good. Thanks!

ConnorDoyle:
Thanks good catch :baseball:

PR #42, opened 2015-04-03
Check for unstaged changes in pre-commit.

### LINKED ISSUE #41: Add a check to the pre-commit hook to make sure there are no unstaged changes.
CC @jsancio

ConnorDoyle:
Fixes #41 

This will cause CI to fail for PRs that don't update the index properly.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -9,3 +9,5 @@ SCRIPTS_DIR=$UNIVERSE_DIR/scripts
 
 $SCRIPTS_DIR/build.sh
 
+# Fail if there are unstaged changes
+git diff --exit-code


hooks/pre-commit
```diff
@@ -9,3 +9,5 @@ SCRIPTS_DIR=$UNIVERSE_DIR/scripts
 
 $SCRIPTS_DIR/build.sh
 
+# Fail if there are unstaged changes
+git diff --exit-code
```
jsancio: This is pretty aggressive but I guess it is okay given how the build command works.

ConnorDoyle: We can make it nicer later, or figure out a different way to update the index.  It's important for now that we ensure that PRs are not breaking it though.  The unfortunate consequence is that PRs have a high potential of needing a rebase in order to be accepted for now.

PR #43, opened 2015-04-03
Update cassandra package version and reduce default consumed resources.

BenWhitehead:


BenWhitehead:
@ConnorDoyle @air please review and merge.

repo/packages/C/cassandra/0/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cassandra",
-  "version": "0.1.0-SNAPSHOT",
+  "version": "0.1.0-SNAPSHOT-396-rewrite-c0e080b4d4",
```
ConnorDoyle: Just FYI, it would be great to not have snapshot builds specified by packages for repeatability.  Is this tag ephemeral, or is the SNAPSHOT part of the version misleading.

ConnorDoyle: OK, synced offline.  This is better that what was there before.

BenWhitehead:
rebased and squashed.

repo/packages/C/cassandra/0/config.json
```diff
@@ -79,19 +79,19 @@
...
       "minimum": 0,
-      "default": 2048
+      "default": 768
     },
     "cassandra/resource/disk": {
       "description": "Disk (MB) to allocate to each Cassandra Server instance.",
       "type": "integer",
       "minimum": 0,
-      "default": 2048
+      "default": 16
```
adam-mesos: Wow, that's really small. But I guess we're not enabling disk isolation (yet), so it will minimize the scheduling/allocation impact and won't actually get enforced.

BenWhitehead: Yup :/

repo/packages/C/cassandra/0/config.json
```diff
@@ -79,19 +79,19 @@
       "description": "CPU shares to allocate to each Cassandra Server Instance.",
       "type": "number",
       "minimum": 0.0,
-      "default": 2.0
+      "default": 0.1
```
adam-mesos: 0.1 Could be too small? Maybe 0.5 or 0.2? Just want to make sure that cpu throttling doesn't slow down C* to the point of the user finding it unusable.

BenWhitehead: @air thoughts?

air: I suspect the average EA cluster will not be doing a lot CPU-wise. Let's test this on RC1 and see what happens.

PR #44, opened 2015-04-03
moved 'git diff --exit-code' from the pre-commit hook to build.sh so tha...

swartzrock:
...t the pre-commit hook AND travis testing will catch unexpected index changes. Should not affect manual builds


swartzrock:
Okay trying again, rebuilt index on master and merged into this branch.

ConnorDoyle:
Superseded by #50.  There's a problem we didn't think about before with this approach, in that build.sh shouldn't assume it lives in a git repo.

PR #47, opened 2015-04-05
Remove unused, improperly named HDFS config options

adam-mesos:
I put these in when I thought we were going to do command-line overrides, but that's been deferred.
Connor since told me that '.'s are not allowed in moustache config names.
Removing these for now. We can add them back properly when needed.


adam-mesos:
@elingg @ConnorDoyle PTAL

ConnorDoyle:
:100: for keeping it simple to start, especially if those values lacked test coverage.

ConnorDoyle:
Synced to `ea`

PR #49, opened 2015-04-05
Require a 'properties' field in package config.json.

ConnorDoyle:
Makes sure the bug fixed in #48 can't happen again.


ConnorDoyle:
cc @swartzrock

PR #57, opened 2015-04-10
Include "tags" in required fields

mgummelt:
It seems "tags" is a required field, so I'm marking it as such.  Also, is "scm" not a required field?  How else does the client find the package?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,10 @@ _Sample package directory layout._
 {
   "name": "foo",
   "version": "1.2.3",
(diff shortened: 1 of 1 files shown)


jsancio:
Looks good!

ConnorDoyle:
Just following up, scm isn't required since it's purely informational and not expected to be used to fetch anything by clients.  This is so that interested folks can have a clue as to where to find the source code, if it's available publicly.

PR #58, opened 2015-04-10
Added dcos history service to the universe

aquamatthias:
@ConnorDoyle please review

I added the dcos history service to the universe.
This is a simple python based service, packaged as docker image.
Is the docker approach ok, or should that package be part of the base image?
The docker repo is currently private with read access for publish2public. Does it has to be a public repo?

Feedback welcome


repo/packages/H/history/0/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "history",
+  "version": "0.0.4",
+  "website": "https://github.com/mesosphere/dcos-history-service",
+  "maintainer": "support@mesosphere.io",
+  "description": "DCOS history application package",
```
ConnorDoyle: Could you expand on this a bit?  Why do users want this?  The description is searchable too.  Also, no need to mention it's a package here :)

repo/packages/H/history/0/package.json
```diff
@@ -0,0 +1,8 @@
...
+  "tags": [ "mesosphere", "dcos", "ui"]
```
ConnorDoyle: Does this program expose a web UI?

repo/packages/H/history/0/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "properties": {},
```
ConnorDoyle: How does the history service find the Mesos master?  We probably need some config here.  Also, how about the port that the REST service binds to?

repo/packages/H/history/0/marathon.json
```diff
@@ -0,0 +1,29 @@
+{
+  "id": "history",
+  "cpus": 0.5,
+  "mem": 512,
+  "instances": 1,
+  "ports": [5055],
```
ConnorDoyle: To require the resource offer to contain this port, you'd also need to set the `requirePorts` field to `true`.

repo/packages/H/history/0/marathon.json
```diff
@@ -0,0 +1,29 @@
+{
+  "id": "history",
+  "cpus": 0.5,
```
ConnorDoyle: Please make this value configurable.

repo/packages/H/history/0/marathon.json
```diff
@@ -0,0 +1,29 @@
...
+  "mem": 512,
```
ConnorDoyle: Please make this value configurable.

repo/packages/H/history/0/marathon.json
```diff
@@ -0,0 +1,29 @@
...
+  "instances": 1,
```
ConnorDoyle: Please make this value configurable.

repo/packages/H/history/0/marathon.json
ConnorDoyle: Alternately, you could switch to BRIDGE networking.

repo/packages/H/history/0/package.json
aquamatthias: No UI provided. It is a dcos dashboard companion.

repo/packages/H/history/0/config.json
aquamatthias: Currently http://leader.mesos:5050 is used. I make this a configuration parameter which defaults to this value.

repo/packages/H/history/0/marathon.json
aquamatthias: The service itself uses whatever port is managed by mesos ($PORT). What is defined here is the service port (haproxy).

repo/packages/H/history/0/marathon.json
aquamatthias: Made it configurable. How can I adjust the values via dcos-cli?

aquamatthias:
@ConnorDoyle 
Incorporated your feedback. 
Made the docker repository public, so everybody can install the package.
Tested the setup on the current EAP setup.

repo/packages/H/history/0/config.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "properties": {
+    "mesos/master/url": {
```
ConnorDoyle: The other frameworks have converged on the key `mesos/url`.  It will be easier for clients to marshal config values that can be shared if packages standardize on some common keys.  This needs to be formalized here in the universe spec, but for now it would be nice to follow that convention here.  See https://github.com/mesosphere/universe/blob/master/repo/packages/M/marathon/0/config.json#L8

aquamatthias: I observed that as well, while looking through other examples. The property here is about the mesos http urls, not the zk urls. To not confuse the user with zk urls, I chose a different key. Any suggestions?

aquamatthias: @ConnorDoyle Any update here? Do you want me to change the property name?

kensipe: the defacto standard here looks like this: https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/M/marathon/1/config.json#L179-L192

repo/packages/H/history/0/config.json
```diff
@@ -0,0 +1,31 @@
...
+      "minimum": 0.0,
+      "default": 0.1
+    },
+    "history/resource/mem": {
+      "description": "Memory (MB) to allocate to each history service instance.",
+      "type": "integer",
+      "minimum": 0,
+      "default": 256
+    },
+    "history/port": {
```
kensipe: also we have a standard of using property hierarchy and dot notation
 and we needed required to be populated.

repo/packages/H/history/0/marathon.json
```diff
@@ -0,0 +1,32 @@
+{
+  "id": "history",
```
kensipe: id value needs to be configurable with a config property

repo/packages/H/history/0/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "history",
+  "version": "0.0.6",
+  "website": "https://github.com/mesosphere/dcos-history-service",
+  "maintainer": "support@mesosphere.io",
+  "description": "This service is a companion service to the dcos dashboard to enable historicized data. It collects the mesos master state of the mesos leader periodically.",
```
kensipe: pre and post notes are recommended.

BenWhitehead:
I this this PR is obsolete and can be closed. Pretty sure the history service it's referencing is shipped in dcos.

aquamatthias:
the history service is part of the dcos base image.

PR #59, opened 2015-04-13
Remove the secret from the dcos wheel

jsancio:
cc @ssk2 

Remove the secret from the helloworld example.


ConnorDoyle:
:shipit:

PR #62, opened 2015-04-14
Point package to correct github repo

elingg:
Point package to correct github repo


elingg:
@jsancio or @ConnorDoyle can you please review at your earliest convenience. thx!

PR #71, opened 2015-04-19
Update HDFS Package for EA2

elingg:
@ConnorDoyle, Please take a look at your earliest convenience, thx! I will also create a similar PR for master once this is merged.

### DIFF
--- a/repo/packages/H/hdfs/0/config.json
+++ b/repo/packages/H/hdfs/0/config.json
@@ -8,7 +8,7 @@
         "type": "string"
       },
       "default": [
-        "https://downloads.mesosphere.io/hdfs/dcos/0.1.0/hdfs-mesos-0.1.0.tgz"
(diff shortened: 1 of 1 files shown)


ConnorDoyle:
For these changes, let's target `master` and then sync to `ea2`.  Trying to make sure we don't get out of sync or lose any important changes in `master`.

repo/packages/H/hdfs/0/config.json
```diff
@@ -8,7 +8,7 @@
         "type": "string"
       },
       "default": [
-        "https://downloads.mesosphere.io/hdfs/dcos/0.1.0/hdfs-mesos-0.1.0.tgz"
+        "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/dcos-ea/0.1.0/hdfs-mesos-0.1.0.tgz"
```
ConnorDoyle: Did the software change, or is it just configuration?  If there were source code modifications we should rev the version instead (doesn't look like a snapshot, so assuming the tag is immutable).

elingg:
Sure, @connorDoyle. Let me do that.

ConnorDoyle:
Thanks!

PR #72, opened 2015-04-19
Update HDFS Package for EA2

elingg:
cc @ConnorDoyle

### DIFF
--- a/repo/packages/H/hdfs/0/config.json
+++ b/repo/packages/H/hdfs/0/config.json
@@ -8,7 +8,7 @@
         "type": "string"
       },
       "default": [
-        "https://downloads.mesosphere.io/hdfs/dcos/0.1.0/hdfs-mesos-0.1.0.tgz"
+        "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/dcos-ea/0.1.0/hdfs-mesos-0.1.0.tgz"
       ]
     }
   }


repo/packages/H/hdfs/0/config.json
```diff
@@ -8,7 +8,7 @@
         "type": "string"
       },
       "default": [
-        "https://downloads.mesosphere.io/hdfs/dcos/0.1.0/hdfs-mesos-0.1.0.tgz"
+        "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/dcos-ea/0.1.0/hdfs-mesos-0.1.0.tgz"
```
ConnorDoyle: Just checking, did the software change, or just configuration? If there were source code modifications we should rev the version instead (doesn't look like a snapshot, so assuming the tag is immutable).

elingg: The updates done were configurations (memory and data directory). We are still at version 0.1.0 (public beta release) for the HDFS Framework.

ConnorDoyle: OK, SGTM.

ConnorDoyle:
Synced to `ea2`

repo/packages/H/hdfs/0/config.json
cmaloney: Why the switch to the non-cloudfront cached URL?

elingg: downloads.mesosphere.io has caching that we want to avoid. We want to point directly to our AWS bucket for that reason.

PR #77, opened 2015-04-21
Use latest Chronos docker container

elingg:
please take a look at your earliest convenience @jsancio or @ConnorDoyle


elingg:
Note, this also needs to be merged into EA2.

elingg:
Hm, not sure why this is so slow running the build. Going to close and reopen.

PR #78, opened 2015-04-21
Update to latest version of chronos

elingg:
@jsancio or @ConnorDoyle, hopefully the build will run ASAP. Please take a look when you have a chance. This also needs to be merged to EA2.


ConnorDoyle:
The version field in package.json needs be updated here as part of this change, since you're now pointing at a new version of Chronos in the target docker image.

elingg:
Thanks for catching that @ConnorDoyle, will do

PR #85, opened 2015-04-22
Update cassandra config and marathon.json to use cluster name to build zk URL unless zk url is specified.

BenWhitehead:


BenWhitehead:
@jsancio please review and merge

PR #88, opened 2015-04-24
Update Chronos default memory usage

elingg:
Update Chronos memory usage so we can successfully process jobs without
out of memory errors


elingg:
@jsancio or @ConnorDoyle, please take a look when you have a chance. This was tested.

repo/packages/C/chronos/0/config.json
```diff
@@ -115,7 +115,7 @@
       "description": "Memory (MB) to allocate to each Chronos instance.",
       "type": "number",
       "minimum": 512.0,
-      "default": 700.0
+      "default": 1024.0
```
ConnorDoyle: Just curious, is this value (or another one) used to set the JVM heap size?

elingg: Heap is actually set separately in our packaging currently. It would be ideal to be able to pass it as well.

ConnorDoyle: OK, could you make an internal JIRA for that?  Otherwise LGTM.

PR #89, opened 2015-04-24
Add basic health check to HDFS

elingg:
This was tested @jsancio or @ConnorDoyle can you take a look please, thx!

### DIFF
--- a/repo/packages/H/hdfs/0/marathon.json
+++ b/repo/packages/H/hdfs/0/marathon.json
@@ -9,7 +9,7 @@
   "healthChecks": [
     {
       "protocol": "COMMAND",
-      "command": { "value": "hadoop fs -ls hdfs://hdfs/ && hadoop fs -rm -f hdfs://hdfs/test.txt && rm -f test1.txt && echo \"this is a test\" > test.txt && hadoop fs -put test.txt hdfs://hdfs/ && hadoop fs -get hdfs://hdfs/test.txt test1.txt" },
+      "command": { "value": "hadoop fs -ls hdfs://hdfs/ && hadoop fs -rm -f hdfs://hdfs/test1.txt && rm -f test2.txt && echo \"this is a test\" > test1.txt && hadoop fs -put test1.txt hdfs://hdfs/ && hadoop fs -get hdfs://hdfs/test1.txt test2.txt" },
       "gracePeriodSeconds": 300,
       "intervalSeconds": 60,
       "timeoutSeconds": 20,
--- a/repo/packages/H/hdfs/0/marathon.json
+++ b/repo/packages/H/hdfs/0/marathon.json
@@ -9,7 +9,7 @@
   "healthChecks": [
     {
       "protocol": "COMMAND",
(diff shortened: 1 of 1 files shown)


repo/packages/H/hdfs/0/marathon.json
```diff
@@ -5,5 +5,15 @@
   "disk": 100,
   "instances": 1,
   "cmd": "{{#hdfs/custom-config}}echo {{hdfs/custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs/custom-config}}cd hdfs-mesos* && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
-  "uris": {{{hdfs/scheduler-uris}}}
+  "uris": {{{hdfs/scheduler-uris}}},
+  "healthChecks": [
+    {
+      "protocol": "COMMAND",
+      "command": { "value": "curl -f -X GET http://namenode2.hdfs.mesos:50070" },
```
ConnorDoyle: Why namenode 2?

BenWhitehead:
:-1: Marathon shouldn't be responsible for checking the health of tasks of another framework.

elingg:
@benwhitehead, we're looking for a quick working solution before creating a full health check. That said, we can run the hdfs commands instead which is the current plan.

elingg:
What are your thoughts @benwhitehead, @connordoyle? this is a test of listing reading and writing, but we could also do a simple ls command.

BenWhitehead:
This seems like a much better test than the curl one from before.

Have you verified that this sequence of commands can succeed on a cluster that is running?

elingg:
@benwhitehead, yes verified!

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -5,5 +5,15 @@
   "disk": 100,
   "instances": 1,
   "cmd": "{{#hdfs/custom-config}}echo {{hdfs/custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs/custom-config}}cd hdfs-mesos* && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
-  "uris": {{{hdfs/scheduler-uris}}}
+  "uris": {{{hdfs/scheduler-uris}}},
+  "healthChecks": [
+    {
+      "protocol": "COMMAND",
+      "command": { "value": "hadoop fs -ls hdfs://hdfs/ && hadoop fs -rm -f hdfs://hdfs/test1.txt && rm -f test2.txt && echo \"this is a test\" > test1.txt && hadoop fs -put test1.txt hdfs://hdfs/ && hadoop fs -get hdfs://hdfs/test1.txt test2.txt" },
```
jsancio: Looks like you are left with test2.txt in the sandbox.

Also, can we at least make the path for the test configurable and document this behaviour? The last thing we want to do is remove our customer's file.

elingg:
read @jsancio's comment. Perhaps to be safe, we should run a simple ls command. That will fail if hfds is not up so that's a reasonable test. Any objections?

ConnorDoyle:
Nice result :+1:

PR #93, opened 2015-04-25
kafka - added basic package

dmitrypekar:
Hey.
Please review initial attempt of dcos kafka package. dcos alpha compatibility. No CLI yet.
Any comments are welcome.


repo/packages/K/kafka/0/marathon.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "kafka",
+  "cpus": 0.5,
+  "mem": 256,
+  "instances": 1,
+  "cmd": "java -jar kafka-mesos*.jar scheduler --master=zk://master.mesos:2181/mesos --zk=master.mesos:2181 --api=http://localhost:$PORT0",
```
tnachen: Is it going to bind on localhost?

dmitrypekar: Nope. It will bind to 0.0.0.0. 
But the URL, that Executor will use to download deps would be incorrect. Need to fix.

dmitrypekar: I've tried using http://master.mesos:$PORT0/ but with no luch. 

For some reason http port of master is not accessible from slave while is it accessible from the outside, using domain name like: http://ec2-54-149-5-171.us-west-2.compute.amazonaws.com:10000 (10000 - assigned port).

What could be causing this?

tnachen: Not sure what's you're trying to do, but $PORT0 is not a port you use to communicate with the master, $PORT0 is what marathon got from Mesos to assign to your framework to use. 
Why do you need to access http://master.mesos:$PORT0/?

dmitrypekar: Scheduler is listening for HTTP connections on that port.
It serves 2 tasks:
- provide REST for CLI;
- distribute binaries for executor;

Of course the executor could be also download from the outside, but, IMHO, it would be better to download it from scheduler, because it is the same jar file and is in local network.

dmitrypekar: I think I've understood. Scheduler is running on any slave in that configuration.

dmitrypekar: Pushed a fix.

tnachen:
Thanks for the patch, just tried out a simple install and run, and it seems to started one broker fine.

PR #95, opened 2015-04-28
kafka package. added values (image, master, zk, etc) to config.json

dmitrypekar:


repo/packages/K/kafka/0/config.json
```diff
@@ -1,6 +1,35 @@
...
+            "type": "string",
+            "description": "--api option",
+            "default": "http://$HOST:$PORT0"
+        },
+        "storage": {
+            "type": "string",
+            "description": "--storage option",
+            "default": "zk:/kafka-mesos"
+        },
+        "other": {
```
tnachen: Calling this other seems pretty confusing, how about calling this optionalFlags or something similiar?

tnachen:
Everything LGTM besides that one config name

jsancio:
How was this tested? Let's make sure that `dcos package install kafka` works without having to supply user options. Thanks!

repo/packages/K/kafka/0/config.json
```diff
@@ -1,6 +1,35 @@
 {
-  "type": "object",
-  "properties": {},
-  "additionalProperties": false,
-  "required": []
+    "type": "object",
+    "properties": {
```
jsancio: Let's make properties "namespaced" take a look at the Marathon package.

repo/packages/K/kafka/0/config.json
```diff
@@ -1,6 +1,35 @@
...
+        "storage": {
+            "type": "string",
+            "description": "--storage option",
+            "default": "zk:/kafka-mesos"
+        },
+        "other": {
+            "type": "string",
+            "description": "other options if required",
+            "default": ""
+        }
```
jsancio: Let's make sure that we add what is required. Looking at the `marathon.json` it looks like we require: master, zk, api, storage and other. [mustache](http://mustache.github.io/mustache.5.html) provides provides mechanisms for dynamically adding text and property resolution.

repo/packages/K/kafka/0/config.json
dmitrypekar: Could you plz, clarify what is the reason for "namespacing" them? We have only 6 properties, so there is no big reason to "structure" them. Beside that, if "namespaced", properties would be called like "kafka/image", "kafka/master", "kafka/api", etc, that, IMHO, would be more confusing to end user then is now.

repo/packages/K/kafka/0/config.json
dmitrypekar: Sure. I've renamed other -> otherOptions. Now it should be self-explainable. Pushed changes.

repo/packages/K/kafka/0/config.json
dmitrypekar: Sure. I've added props for all those settings in config.json and tested that they work on my test dcos cluster.

repo/packages/K/kafka/0/config.json
jsancio: Namespacing is not required but it is nice because users can store uber user options files that are specialized for their datacenters.

tnachen:
I tried it myself already and didn't have to provide options

repo/packages/K/kafka/0/config.json
dmitrypekar: So, do you suggest to add "kafka/" namespace, so config properties would be called, for instance, 
kafka/master, kafka/zk, etc? If you think, that this approach is pragmatic, I will add those namespace.

jsancio: Yes. You can use "kafka/". We are also in the process of improving the Marathon package so that we can use that as an example for other packages to follow. See: https://github.com/mesosphere/universe/pull/102

dmitrypekar: Fixed, tested, pushed. Now every prop in config.json is prefixed with "kafka/"

PR #96, opened 2015-04-28
Fix python image name/tag for helloworld.

adam-mesos:
Needed to change the '/' to a ':' to use the tag properly.
Also switched to just `:3` so we automatically get the latest 3.x.y
@jsancio PTAL

### DIFF
--- a/repo/packages/H/helloworld/0/marathon.json
+++ b/repo/packages/H/helloworld/0/marathon.json
@@ -7,7 +7,7 @@
   "container": {
     "type": "DOCKER",
     "docker": {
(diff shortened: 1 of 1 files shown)


jsancio:
@adam-mesos, do we need this in EA2?

adam-mesos:
@jsancio Do we tell people to run helloworld? If not, then we can leave it out.
Regardless, it's a trivial fix, so might as well merge it. Doubt that it should cause a new DCOS build/version itself.

jsancio:
I don't think so. I will merge to master branch. EA2 is on the ea2 branch.

PR #106, opened 2015-05-04
Update cassandra config to use new nested object support.

BenWhitehead:


BenWhitehead:
@ConnorDoyle or @jsancio Please review and merge

repo/packages/C/cassandra/0/config.json
```diff
@@ -1,104 +1,135 @@
...
+              "type": "integer",
+              "minimum": 0,
+              "default": 604800
+            },
+            "role": {
+              "description": "Mesos role for this framework.",
+              "type": "string",
+              "default": "*"
+            },
+            "authentication": {
```
ConnorDoyle: I'd recommend setting `additionalProperties` to false, at least for this nested object.  That will make typos fail the validation step.  Imagine you had mis-spelled `enabled`...

repo/packages/C/cassandra/0/config.json
```diff
@@ -1,104 +1,135 @@
...
+              "default": 768
+            },
+            "disk": {
+              "description": "Disk (MB) to allocate to each Cassandra Server instance.",
+              "type": "integer",
+              "minimum": 0,
+              "default": 16
+            },
+            "heap-mb": {
+              "description": "The amount of memory in MB that are allocated to each Cassandra Server Instance. This value should be smaller than 'cassandra/resource/mem' the remaining difference will be used for memory mapped files and other off-heap memory requirements.",
```
ConnorDoyle: Run-on sentence (break before "the remaining difference..."?).

repo/packages/C/cassandra/0/config.json
```diff
@@ -1,104 +1,135 @@
...
+          "description": "The minimum number of seconds to wait between starting each node. Setting this too low could result in the ring not bootstrapping correctly.",
+          "type": "integer",
+          "minimum": 15,
+          "default": 120
+        },
+        "data-directory": {
+          "description": "The location on disk where Cassandra will be configured to write it's data.",
+          "type": "string",
+          "default": "."
+        },
```
ConnorDoyle: extra whitespace: intentional?

ConnorDoyle:
Tested on DCOS with the default values?

repo/packages/C/cassandra/0/config.json
jsancio: Hmm. If we want to support an uber options file for the whole data center then we shouldn't have additional properties set to false. Thoughts?

jsancio: Ignore my comment. I think inside the `cassandra` object it should be okay. :+1: for hierarchical config files.

ConnorDoyle: Exactly :guitar:

BenWhitehead:
@ConnorDoyle updated based on feedback.  Yes, I've tested it on DCOS with default values and custom values.

ConnorDoyle:
:shipit:

PR #107, opened 2015-05-05
Remove the sample file

jsancio:
Let's remove all of the sample file since they are not technically part
of the spec. We can add it later to the universe specification if we
need this.

cc @BenWhitehead @swartzrock


ConnorDoyle:
Goodbye inconsistent and unspecified repository content!  See you never.

PR #110, opened 2015-05-07
Hue

abhay-agarwal:


abhay-agarwal:
@jsancio can you please look over? I'll actually merge once the requisite changes have been made to HDFS.

jsancio:
High-level note. Please make this package configurable.

abhay-agarwal:
Okay. Which parameters should be configurable?

abhay-agarwal:
updated @jsancio

elingg:
Thanks @abhay-agarwal! Before merging we need to point the following URL here, https://github.com/mesosphere/universe/blob/master/repo/packages/H/hdfs/0/config.json#L11: https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz

abhay-agarwal:
@jsancio This is good to ship!

repo/packages/H/hue/0/config.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "object",
+  "properties": {
+    "hue/configuration": {
```
jsancio: We now recommend people to use hierarchical options files. See Marathon and Cassandra. Thanks!

elingg:
hi @abhay-agarwal, I just caught this, too. We also need to update the version to 0.1.1 here: https://github.com/mesosphere/universe/blob/master/repo/packages/H/hdfs/0/package.json#L3. With that change we also need to rebuild the index.

abhay-agarwal:
Good to :ship: @jsancio

abhay-agarwal:
I also updated the version number of HDFS and updated the HDFS subcommand as per Connor's comments.

abhay-agarwal:
@jsancio Can I merge this?

repo/packages/H/hue/0/marathon.json
```diff
@@ -0,0 +1,52 @@
...
+                "hostPath": "/opt/mesosphere/etc/hadoop/",
+                "mode": "RO"
+            }
+        ]
+    },
+    "uris": [
+        "{{hue.configuration}}",
+        "{{httpfs.configuration}}"
+    ],
+    "cmd": "export PATH=/opt/mesosphere/bin:$PATH && \\cp -f /etc/hadoop/conf/*-site.xml /etc/hadoop-httpfs/conf/ && \\cp -f $MESOS_SANDBOX/httpfs-site.xml /etc/hadoop-httpfs/conf/ && service hadoop-httpfs start; \\cp -f $MESOS_SANDBOX/hue.ini /etc/hue/conf/hue.ini && /usr/lib/hue/build/env/bin/hue runserver 0.0.0.0:8000",
```
jsancio: This is PATH manipulation looks very DCOS Image specific. Is this so we can get access to `service`? The user should be able to overwrite this to work on their env. cc @cmaloney

abhay-agarwal: True. Let me fix.

abhay-agarwal:
@jsancio I have updated this package. Take a look

repo/packages/H/hue/0/marathon.json
```diff
@@ -0,0 +1,53 @@
...
+                "containerPath": "/etc/hadoop/conf/",
+                "hostPath": "{{hadoop.conf-dir}}",
+                "mode": "RO"
+            }
+        ]
+    },
+    "uris": [
+        "{{hue.configuration}}",
+        "{{httpfs.configuration}}"
+    ],
```
elingg: hi @abhay-agarwal, for this cmd, what you should do is directly download the Hadoop package that you need instead of mounting directories in DCOS image as volumes. We want to avoid dependencies on DCOS Image if possible, and it should be possible here. If you have questions about this, feel free to discuss with @cmaloney or ask me.

elingg:
Thanks @abhay-agarwal, I took a look and left a comment that should be addressed. Once that's taken care of, we can do a final review.

abhay-agarwal:
@elingg our system architecture is highly dependent on the hadoop that is actually at `/opt/mesosphere/bin/` because it references other configs. Similarly, I have to mount the entire `/opt/mesosphere/` folder because there are symlinks that are referenced throughout. Until we develop a better architecture for Hadoop, this is the best solution I can engineer.

cmaloney:
I don't see a good reason why you couldn't download the hdfs, static config inside of the Hue app itself rather then relying on the copy of (more or less) the exact same in the host system. What is structurally so hard to download / add to path yourself?

elingg:
hi @abhay-agarwal, +1 to @cmaloney's comment. Just download the hdfs tarball and use that instead of depending on DCOS Image. It has the config preset. That would be much cleaner.

abhay-agarwal:
@elingg @cmaloney The only drawback there is that i'll have to update this package every time our HDFS tarball changes. Which could potentially be every week/month/etc.

elingg:
@abhay-agarwal, that's okay, we just need to update the package to link to the latest released version of HDFS. We need to do this anyway in Universe each time there is a release.

cmaloney:
If you use the base system ones:

1. You are getting a special exception from the rules (Don't depend on the host system), your service isn't self contained.
2. The system packages are going to change without warning / notice to your DCOS Service. This likely means at some point there will be a system update independent of your application update which would completely break Hue without notice to users. If things are self-contained you'll at least get "files exist" and hopefully some reasonable error messages about incompatible network APIs / protocol versions.

Practically this all depends on the hdfs-mesos framework, not the host HDFS for functionality. I think there is a way to specify that dependency.

abhay-agarwal:
Agreed. I made the change in the packge so there's no more volume mounting. Would be great to have a dependency system in-place by the next time I have to make the update :shipit:

elingg:
Yes, I think this is the best way to include this dependency is linking to it as opposed to using the base system. I will take a look at this PR ASAP.

repo/packages/H/hue/0/config.json
```diff
@@ -0,0 +1,66 @@
+{
+  "type": "object",
+  "properties": {
+    "hdfs": {
+      "description": "Hadoop specific configuration properties",
+      "type": "object",
+      "properties": {
+        "scheduler-uri": {
```
elingg: nit: In description, it should say (must being with 'hdfs-mesos') instead of  "description": "HDFS package artifact tarball (must begin with 'hdfs')"

elingg: Instead of calling this scheduler URI, this should be called package URI.

repo/packages/H/hue/0/config.json
```diff
@@ -0,0 +1,66 @@
...
+    "httpfs": {
+      "description": "Httpfs specific configuration properties",
+      "type": "object",
+      "properties": {
+        "configuration": {
+          "description": "A custom universally-accessible URL of a httpfs-site.xml file.",
+          "type": "string",
+          "default": "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/hue/httpfs-site.xml"
+        },
+        "port": {
```
elingg: The description needs to be updated here.

repo/packages/H/hue/0/config.json
```diff
@@ -0,0 +1,66 @@
...
+    },
+    "httpfs": {
+      "description": "Httpfs specific configuration properties",
+      "type": "object",
+      "properties": {
+        "configuration": {
+          "description": "A custom universally-accessible URL of a httpfs-site.xml file.",
+          "type": "string",
+          "default": "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/hue/httpfs-site.xml"
+        },
```
elingg: Quick question, why do we need port for "httpfs" and for "hue"?

abhay-agarwal: This hue package bundles a httpfs server (its the only way to address an HA hdfs install). Currently, both are served via port 0, but if a user wants a specific port for either (to support some other tool, etc) they can.

elingg: makes sense, thanks

repo/packages/H/hue/0/config.json
```diff
@@ -0,0 +1,66 @@
...
+        "scheduler-uri"
+      ]
+    },
+    "hue": {
+      "description": "Hue specific configuration properties",
+      "type": "object",
+      "properties": {
+        "configuration": {
+          "description": "A custom universally-accessible URL of a hue.ini file.",
+          "type": "string",
```
elingg: Let's move the bucket for Hue outside of HDFS. Remember that it will be used by Myriad, Spark, etc so it is a separate entity than HDFS.

abhay-agarwal: I don't follow. Do you mean remove the tag in `package.json`?

repo/packages/H/hue/0/config.json
elingg: nit: In description, it should say (must being with 'hdfs-mesos') instead of "description": "HDFS package artifact tarball (must begin with 'hdfs')"

repo/packages/H/hue/0/config.json
elingg: Let's make this "https://s3.amazonaws.com/downloads.mesosphere.io/hue/httpfs-site.xml" instead of "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/hue/httpfs-site.xml"

abhay-agarwal:
@elingg updated with your changes.

elingg:
LGTM! Any objections @jsancio or @cmaloney? If not, we should be ready to merge.

repo/packages/H/hdfs/0/command.json
```diff
@@ -1,5 +1,5 @@
 {
   "pip": [
-    "git+https://github.com/mesosphere/dcos-hdfs.git@0.1.0#dcos-hdfs=0.1.0"
+    "git+https://github.com/mesosphere/dcos-hdfs.git@0.1.1#dcos-hdfs=0.1.1"
```
jsancio: Do we have plans to remove this technical debt?

cmaloney:
looks reasonable to me from a "doesn't touch the base image" perspective.

repo/packages/H/hdfs/0/command.json
abhay-agarwal: I'm still not sure how we would send a local configuration file without encoding it.

jsancio: I meant that we should be using wheels to package and install Python packages. We should be setting standard for others to follow.

elingg: +1 this would be nice to get rid of. I'm not familiar with the reasoning behind creating another repo, but I would like to remove this repo if we can. What course of action do you suggest @jsancio?

abhay-agarwal: but if you mean rewriting the hdfs universe package to take different config parameters, its not on the roadmap yet.

abhay-agarwal: The separate repo is how Universe packages install CLI tools. See how spark does it. If you remember from around a month ago, we had a push to 'make hdfs configurable' so I implemented the only solution that was within the time constraints. Since we have another three weeks now, we can add a better solution to the roadmap.

abhay-agarwal: In the meantime, this update fixes https://mesosphere.atlassian.net/browse/DCOS-1093

elingg: Thanks for clarifying, @abhay-agarwal. So the separate repo is indeed necessary, but the way we access it in Universe is not ideal.  We can merge this now and then @abhay-agarwal could add a JIRA to fix this technical debt and work on it within the next 3 weeks. Does that sound like a good plan, @abhay-agarwal and @jsancio?

abhay-agarwal: https://mesosphere.atlassian.net/browse/HDFS-85

elingg: Thanks. Actually I thought about it and configuring it would not be hard using options.json. We'd just ask the users to point to a URL with their config file, which is the common way things are handled in Universe. This would be a fairly quick fix, right, or am I missing something?

abhay-agarwal: that fix is already available via 'scheduler-uris' in hdfs package. This is to allow a user to specify individual parameters in options.json.

elingg: Ok, discussed with others and for passing the config, the proper fix is https://mesosphere.atlassian.net/browse/HDFS-25. I closed HDFS-85 for this reason.

The main issue that @jsancio was referring to was a team city build and an automated update to an s3 bucket for the subcommand code: git+https://github.com/mesosphere/dcos-hdfs.git@0.1.1#dcos-hdfs=0.1.1. Since automated builds are important and hdfs itself is now built in team city, could you do this?

abhay-agarwal: Can we move that to a separate PR? This is not an omnibus update to HDFS.

elingg: Sure, fine by me. Let's create a JIRA to automate the build of the dcos-hdfs (hdfs subcommand). Will you be looking into it? If so, assign it to yourself. If not, assign it to me for now.

elingg: Cool, thanks for creating the JIRA and assigning to me. We should be good to go!

elingg:
Ship It! :shipit:

PR #118, opened 2015-05-13
Helloworld

mgummelt:
remove docker so we can run this in TC (there were docker on docker issues)


mgummelt:
I tested on DCOS

ConnorDoyle:
:ship: it.

cmaloney:
This relies on the host python3 in DCOS, which is strictly against the Service guidelines. Please don't do that! The python is not guaranteed to be there and will likely move without notice).

mgummelt:
can you link me to the service guidelines?

cmaloney:
Not shared publicly at the moment I don't think. Sending to you privately.

PR #119, opened 2015-05-13
Revert helloworld

mgummelt:
reverting previous change, since we can't assume python is installed on DCOS hosts


ConnorDoyle:
Thanks for noting the problem @cmaloney.  Un:ship: it.

PR #121, opened 2015-05-15
Update DCOS_PACKAGE_FRAMEWORK_NAME for Chronos and HDFS

elingg:
Please take a look, @jsancio and @ConnorDoyle or anyone else available to review!


repo/packages/C/chronos/0/marathon.json
```diff
@@ -28,6 +28,8 @@
       "portIndex": 0,
       "protocol": "HTTP"
     }
-  ]
-
+  ],
+  "labels": {
+    "DCOS_PACKAGE_FRAMEWORK_NAME": "{{chronos/mesos-framework-name}}"
```
jsancio: If we are going to do this then we need to make this a required property.

jsancio: FYI, Ideally we shouldn't expose this label stuff in the Marathon JSON. It just happen that right now we are using Marathon to store this information. This is a short term solution and very implementation specific. We should try to mimic how the Marathon package does it by using hierarchical config.json templates and a well known name.

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -15,5 +15,8 @@
       "timeoutSeconds": 20,
       "maxConsecutiveFailures": 3
     }
-  ]
+  ],
+  "labels": {
+    "DCOS_PACKAGE_FRAMEWORK_NAME": "hdfs"
```
jsancio: See my comments in the Chronos package.

jsancio:
When are we planning to change this package to use hierarchical config.json schema? See how the Marathon and Cassandra package do this.

repo/packages/C/chronos/0/marathon.json
elingg: It has a default value actually so it's taken care of just like {{mesos/master}} and  {{chronos/zk-hosts}}. That said, I realize I need to update the command to expect that it exists. Just updated!

elingg:
Yes, agreed, we want our services to be model citiziens. There are JIRAs open to change to the hierarchical config.json schema.

elingg:
Please take a second look when you have a chance, @jsancio.

repo/packages/C/chronos/0/marathon.json
ConnorDoyle: Please don't set these labels in the package.  These labels are managed by the DCOS package CLI.  There shouldn't be anything that specifically says "DCOS" in the universe.

ConnorDoyle: e.g. if we insist on a well-known `<package-name>.framework-name` config value for packages where `framework == true`, then the package manager can store this info as necessary.

elingg: @ConnorDoyle, I was using Cassandra as an example. It seems to be the current workaround we are using? https://github.com/mesosphere/universe/blob/master/repo/packages/C/cassandra/0/marathon.json#:L34:L36 cc @BenWhitehead

ConnorDoyle: IMO it should not be there either.

ConnorDoyle: We had an internal JIRA for how this was supposed to work (DCOS-774).  It's been closed, maybe @jsancio has more details about decisions that were made after I stopped tracking it :)

elingg: Ok, thanks Connor. I will wait until he takes a look.

jsancio: Yes, ideally Cassandra shouldn't be doing that either. Unfortunately currently Cassandra generate the framework name from other information. In this case the cluster name.

Having said that, this is not the case for Chronos. Chronos doesn't generate the framework name. It is instead configurable in a very similar fashion to Marathon. We should use the same mechanism that Marathon uses.

elingg: Let me talk to you in person, @jsancio. Chronos is like Cassandra in that it generates the framework name from other information (a config variable called mesos framework name).

elingg:
Hello @jsancio or @ConnorDoyle, I'm pleased to use the hierarchical config schema and remove this technical debt! Now these packages can be model citizens. Please take a look.

BenWhitehead:
Cassandra is a little different than chronos in that framework name itself isn't directly configured. What is configurable is the cassandra cluster name.

Cluster names are what has to change between cassandra clusters, as such the cassandra framework absorbs the logic to ensure that a different marathon api id and framework id are always generated for the cluster.

The reason cassandra specifies `DCOS_PACKAGE_FRAMEWORK_NAME` is so that cassandra can tell the CLI and other tools the framework name that will be registered.

repo/packages/C/chronos/0/config.json
```diff
@@ -1,199 +1,229 @@
...
+        "type": "string"
+      }
+    }, 
+    "required": [
+      "cpus",
+      "instances",
+      "hostname",
+      "id",
+      "mem",
+      "zk-hosts"
```
jsancio: We should add `framework-name` as a require property.

jsancio:
Looks good to me. One minor comment.

repo/packages/C/chronos/0/config.json
elingg: Agreed. Done!

elingg:
Thanks @jsancico, I made the fix based on your comment. I will merge now, unless there are any final comments.

jsancio:
Squash the commits and merge. Thanks!

PR #124, opened 2015-05-19
kafka_dcos_compat_changes

dmitrypekar:
Hi,
I've just implemented following changes. Please merge.

The Items that are critical to get done are:
1. Health Checks
  1. Add health check for scheduler process that is independent from cluster health
  2. Set maxConsecutiveFailures to 0 for cluster healthcheck

2. Framework name
  1. Add config parameter to config.json in universe
  2. Reference config parameter from config.json in marathon.json

3. Universe config.json
  1. Switch to using the hierarchical definition format used for namespacing
  2. Define all properties listed as required regardless of a default value being provided
    1. Replace kafka/master with mesos.master

  3. Set `additionalProperties` to false for kafka properties object

4. Support setting the role on FrameworkInfo at registration

5. Support setting the user on FrameworkInfo at registration

6. Support framework authentication

7. Update `postInstallNotes` in package.json to include a link to framework documentation/cli documentation

8. Update `postInstallNotes` in package.json to include a link to where issues can be reported

### DIFF
--- a/repo/packages/K/kafka/0/marathon.json
+++ b/repo/packages/K/kafka/0/marathon.json
@@ -14,6 +14,15 @@
       "intervalSeconds": 60,
       "portIndex": 0,
       "timeoutSeconds": 30,
+      "maxConsecutiveFailures": 3
+    },
+    {
+      "protocol": "HTTP",
+      "path": "/api/status",
+      "gracePeriodSeconds": 120,
+      "intervalSeconds": 60,
+      "portIndex": 0,
+      "timeoutSeconds": 30,
       "maxConsecutiveFailures": 0
     }
   ]
--- a/repo/packages/K/kafka/0/marathon.json
+++ b/repo/packages/K/kafka/0/marathon.json
@@ -3,7 +3,7 @@
   "cpus": 0.5,
   "mem": 256,
   "instances": 1,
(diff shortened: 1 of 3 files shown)


repo/packages/K/kafka/0/marathon.json
```diff
@@ -3,18 +3,18 @@
...
       "protocol": "HTTP",
-      "path": "/api/brokers/status",
+      "path": "/health",
       "gracePeriodSeconds": 120,
       "intervalSeconds": 60,
       "portIndex": 0,
       "timeoutSeconds": 30,
-      "maxConsecutiveFailures": 3
+      "maxConsecutiveFailures": 0
     }
```
BenWhitehead: Both health checks can be listed here with the correct parameters so that when the cluster is unhealthy marathon doesn't try and restart the scheduler, but when the scheduler is unhealthy marathon can restart it.

dmitrypekar: Hi Ben.
Actually the current healthcheck only checks Scheduler's process health. If the process is alive it will respond. If not - it will not. We do not have a dedicated healthcheck for cluster health. And in terms of Scheduler it is hard to define such criterion. So for now there is only process healthcheck (for the Scheduler process itself), which could be included in marathon.json. It is served by /health URL.

dmitrypekar: Just added second health check for cluster status with maxConsecutiveFailures=0, changed health check for Scheduler process to have maxConsecutiveFailures=3

BenWhitehead:
Looks good, thanks.

PR #129, opened 2015-05-21
Define Spec for Universe elements

BenWhitehead:
To view rendered doc: https://github.com/mesosphere/universe/blob/spec-doc/docs/spec.md

Layout feedback welcome, I'm not sure about the headers being used right now.

/cc @air @ConnorDoyle @jsancio

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1,0 +1,114 @@
+Universe Spec
+=============
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## MUST define a `package.json`
+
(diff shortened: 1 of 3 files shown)


docs/spec.md
```diff
@@ -0,0 +1,114 @@
...
+
+Schema: [config-schema.json](../repo/meta/schema/config-schema.json)
+
+### MUST namespace all properties
+
+*A single `options.json` should be able to be used for an entire DCOS Cluster*
+
+### MUST use hierarchical definition format
+
+1. MUST set `additionalProperties` to `false` for all nested objects
```
jsancio: Not all nested objects. I think we want this only on the package specific JSON object. For example I don't think we want it to be `false` for the Mesos' nested object.

jsancio:
Very nice. Looks good to me.

jsancio:
@BenWhitehead Should I merge this? Is this good to go?

BenWhitehead:
Nope, not ready to merge yet. This can hold off for a bit.

jsancio:
Sounds good. I added WIP label.

PR #132, opened 2015-05-22
Update Marathon JSON with respect to ports

elingg:


elingg:
Please take a look @jsancio when you have a moment thanks!

jsancio:
Looks good to me. It looks like Chronos support configuring external services endpoints. E.g. cassandra, graphite, etc. How do we plan to discover those? Can we use Mesos DNS? Thanks.

elingg:
@jsancio, yes, we use Mesos DNS for this, i.e. cassandra-dcos-node.cassandra.dcos.mesos for the cassandra hostname. We tested this out with DCOS recently.

PR #133, opened 2015-05-26
Update mem/cpu and add postInstall notes.

aquamatthias:
@ConnorDoyle Please Review


air:
Connor is out, assigning to Jose for merge

aquamatthias:
Hey @jsancio I see the teamcity build is failing. The index.json was changed by the build script `scripts/build.sh`. Any idea what the problem is?

jsancio:
> Hey @jsancio I see the teamcity build is failing. The index.json was changed by the build script scripts/build.sh. Any idea what the problem is?

Try running `scripts/build.sh` locally and committing the index.json along with your PR. I think TC fails the build if running `scripts/build.sh` leads to files changing.

aquamatthias:
As I said: did exactly that, which changed the file on my machine.

jsancio:
Hmm. Interesting. I checked out the branch and ran the command locally and got the same result from TC. I haven't looked at that code in a while but I suspect outdated version of some library.

repo/meta/index.json
```diff
@@ -1,6 +1,9 @@
 {
   "packages":[
     {
```
elingg: Why do we need this addition to index.json? This could be what is breaking the build.

elingg:
hi @aquamatthias, left a note about why this may not be building. Reassigning to you to investigate.

aquamatthias:
@jsancio @elingg The build script creates wrong index.json inde.json.gz.
Removed the change manually.

elingg:
:+1: , any objections @jsancio?

repo/packages/M/marathon/0/config.json
```diff
@@ -14,14 +14,14 @@
           "type": "boolean"
         },
         "cpus": {
-          "default": 1.0,
+          "default": 2.0,
```
mgummelt: Any reason for this?

repo/packages/M/marathon/0/config.json
```diff
@@ -93,9 +93,21 @@
           "type": "integer"
         },
         "mem": {
-          "default": 512.0,
-          "description": "Memory (MB) to allocate to each Marathon instance.",
-          "minimum": 0.0,
+          "default": 1024.0,
```
mgummelt: Any reason for this?

repo/packages/M/marathon/0/package.json
```diff
@@ -10,5 +10,6 @@
     "icon-medium": "https://downloads.mesosphere.io/marathon/assets/icon-service-marathon-medium.png",
     "icon-large": "https://downloads.mesosphere.io/marathon/assets/icon-service-marathon-large.png"
   },
-  "tags": ["mesosphere", "framework"]
+  "tags": ["mesosphere", "framework"],
+  "postInstallNotes": "Marathon DCOS Service has been successfully installed!\nWe recommend a minimum of one node with at least 2 CPU's and 1GB of RAM available for the Marathon Service.\n\n\tDocumentation: https://github.com/mesosphere/marathon\n\tIssues: https:/github.com/mesosphere/marathon/issues\n"
```
mgummelt: add another slash in the protocol...lol

And I'll reiterate that I don't like these resource requirement notes.  Resource failure messages should be an automated part of the system.

mgummelt:
I'm unfamiliar with any build or index issues that might be going on.  We might want @jsancio to comment on that.

But after my suggested changes, this LGTM

repo/packages/M/marathon/0/package.json
elingg: Agreed, this is a temporary fix, but this is what we need to do for now so the user is at least informed.

elingg:
@aquamatthias, could you comment on the increased resource requirements? Other than that, we should be good to merge. Thanks!

repo/packages/M/marathon/0/package.json
aquamatthias: I also think, that this values should come from the proposed default settings.

aquamatthias:
@elingg @mgummelt 
Regarding resource requirements:
1) I did not create the original values in the first place. If the author can make sure it works with these settings - I am ok with it.
2) The settings highly depend on the planned installation: nr of apps, tasks, health checks etc.
3) A CPU of one is very unfortunate, since the default parallelism will rely on the # of cores, which is probably greater than 1. 512mb for the whole process incl Java Heap (and also unrestricted!) will lead to problems and kills from mesos.
4)I understand the values as recommendation for the average use case, not as minimum requirements.

ConnorDoyle:
@aquamatthias: Regarding 3 above, the CPU allocation is translated to cpushares, and the JVM will still see the correct number of cores, but why is this a problem?  The scheduler process can still run on multiple cores concurrently.

aquamatthias:
@ConnorDoyle the nr of cores will translate to the default parallelism. But the effective parallelism is 2 (or would be 1 in the case of one cpu).

mgummelt:
The github http typo was never fixed

ConnorDoyle:
@aquamatthias I was just confused since you seemed to be implying that setting cpushares would pin the process to one core, which it does not.

PR #141, opened 2015-06-02
Add licenses property to package-schema.json

BenWhitehead:
This change allows for a package to list the license(s) of the project.

PR #143, opened 2015-06-02
Add license definition for cassandra

BenWhitehead:


BenWhitehead:
Build will fail until #141 is merged.

PR #149, opened 2015-06-02
Official K8s icons

abhay-agarwal:
Small change that was annoying me.

### DIFF
--- a/repo/packages/K/kubernetes/0/package.json
+++ b/repo/packages/K/kubernetes/0/package.json
@@ -6,9 +6,9 @@
   "description": "Manage a cluster of Linux containers as a single system to accelerate Dev and simplify Ops.",
   "framework": true,
   "images": {
-    "icon-small": "https://downloads.mesosphere.io/kubernetes/artifacts/images/k8s-48x48.png",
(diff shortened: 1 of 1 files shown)


cmaloney:
Why are we changing the url here? downloads.mesosphere.com is setup so that things will leave the cache over time automatically.

abhay-agarwal:
This is more transparent than me changing random s3 files.

cmaloney:
Not particularly... I have no idea what the old vs. new is, they still could change without warning / notice.

The only way this would get particularly transparent is if the icons live inside the git repository and are automatically updated / deployed on commit (Or are grabbed from github directly)

abhay-agarwal:
Okay i'll change them then.

jdef:
There is a build job in TC that uploads icons from the kubernetes-mesos
github repo.

abhay-agarwal:
Our UI team has official icons that don't have the ugly black background and scale better to retina screens. I'm not sure why there needs to be that degree of automation to create a team city job but whatever makes people happy. I'll stay out of it.

jdef:
we don't need that level of automation. and I like the new icons. i only
mentioned the automation because i don't have direct access to the s3
bucket and the only way for me to get icons up there was via the build
process, as recommended by cody, et al.

PR #151, opened 2015-06-03
Update spark preinstall notes

tnachen:


tnachen:
@air is this what you like?

PR #155, opened 2015-06-03
Added validation for repo schema version.

ConnorDoyle:
cc @jsancio


BenWhitehead:
Looks great! Merging

repo/meta/index.json
```diff
@@ -112,5 +112,5 @@
       }
     }
   ],
-  "version":"0.1.0-alpha"
+  "version":"1.0.0-rc1"
```
mgummelt: Isn't this file auto-generated?  If so, where is this version coming from?

mgummelt: It looks like there's a repo/meta/version.json file that's still 0.1.0.  Is that a mistake?

mgummelt:
why push this into a non-master branch?

BenWhitehead:
@mgummelt master has been replaced by `version-1.x` as the default branch.  The intention behind this is that anything in the `version-1.x` branch is compatible with the version 1 schemas.

When we move to version 2 schemas in the future those will be targeted at different branches. This is to allow the cli to validate the repo's version so that it can ensure that it will be able to perform the actions it knows about.

mgummelt:
yea, my mistake for not reading the eng email.  thanks

PR #156, opened 2015-06-04
fix typo Cassandra -> Kafka in Kafka package.json

lloesche:
typo fix


BenWhitehead:
Good catch @lloesche !

PR #161, opened 2015-06-11
Added a config for the MemSQL Mesosphere framework

wsong:


BenWhitehead:
:eyes:

tnachen:
Is this pending review or?

kensipe:
@tnachen yes... I am review

kensipe:
@wsong email sent with details.   There is a blocking issue with the web ui.  It has absolute (not relative) paths which doesn't allow the UI to work from the service rewrite url path.

kensipe:
@wsong the latest docker image for the scheduler keeps failing on marathon install.  Is this ready?  Have you tested this on dcos?

wsong:
It works fine for me.  Are you seeing any specific error messages?

kensipe:
@wsong the build has failed... any ideas?

wsong:
It looks like a transient error with your guys's Docker build system:

https://teamcity.mesosphere.io/viewLog.html?buildId=24014&buildTypeId=Oss_Universe_Ci&tab=buildLog

Can you run it again?

kensipe:
@wsong is it otherwise ready to go?

wsong:
If it looks good to you, then we're ready, yeah.

repo/meta/index.json
```diff
@@ -98,6 +98,20 @@
       }
     },
     {
+      "currentVersion":"0.0.1",
+      "description":"MemSQL running on Apache Mesos. This framework provides the ability to create and manage a set of MemSQL clusters, each running with the MemSQL Ops management tool.",
```
kensipe: this looks great thx!

kensipe:
@wsong I will be following up on this today

wsong:
Thanks Ken.  One issue that I do want to resolve before we ship this is the Amazon internal IPs issue; our web UI uses the offer.hostname field on the Offer protobuf to figure out what host our primary node is running on:

https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L773

But on AWS, this seems to be set to an internal Amazon hostname, which makes our web UI not very useful (since we'll display a link with an internal Amazon hostname, which users can't click on).  You've mentioned this as an issue in the past; how do other frameworks work around this?  Are there any examples we can look at?

kensipe:
@wsong the slaves are all private.  Even if you had a public IP we would block it.   If you are providing a service to services in the cluster the private IP will work fine.  If you are looking to have access external to the cluster then... you will need to have a routing service to a public slave.

if you are looking for a router example, we have a project called the edgerouter:  https://github.com/mesosphere/edgerouter

Our Oinker example app uses a version of it.   What is your intention?   

If you are looking to provide connection details for cluster deployed apps, then we are fine.
If you are looking to have users explore the tutorial, then I would like to see the web ui proxy the calls to the cluster.
if you are looking to allow external connectivity then you will likely need an edge router or proxy of some kind.

wsong:
Ken,

MemSQL is a database, so we need some way for users to be able to connect to the database.  The only node that users have to be able to reach is the master aggregator, but if a slave fails, then we'll automatically promote a different node to be the master aggregator, so the user needs to be able to potentially talk to all slaves.

kensipe:
@wsong of course :)  I'm trying to address your question.  The internal IP / host names are fine for a service that is in the cluster trying to connect to the DB.  Are you looking to allow people to connect from outside the cluster?  if so then there is some work to do.

kensipe:
would it help to get on a call regarding this?

wsong:
Ken,

Yeah, that's correct; our intention is that users will be able to connect to the database not only from within the DCOS cluster, but also from outside it.

kensipe:
in order to connect to it external to the cluster, you will need either:
1) deployment to a public slave
2) a proxy on the public slave to route to the cluster.

kensipe:
Let's get this passing and certified for internal consumption.  Lets consider the external connection as another feature for a later PR.  Does that work for you?

wsong:
Works for me.

BenWhitehead:
I would recommend to not expose the master aggregator outside the cluster by default, this is the same approach that has been taken by cassandra and people are generally fairly comfortable with it.

The one thing we have done to allow people to play with the db, is provide instructions of how to ssh into the cluster then start a docker container with the cassandra toolset in it.

kensipe:
@BenWhitehead if we do want to expose the db to clients external to the cluster, then constraining the aggregator to the public slave would be the preferred way correct?  even if it isn't the "default" option of installation.

BenWhitehead:
Public slaves tend to be much fewer in number, so it would probably be better to launch a "proxy connection" service on the public slave and leave the master aggregator on the normal slaves.

I don't know what sort of protocol is used by memsql, but I assume it's a tcp connection which could be handled by a docker container running HAProxy that is configured by the scheduler.

kensipe:
@wsong missing from the configuration is the ability to change the role.  
example config: https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/M/marathon/1/config.json#L121
example use:  https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/M/marathon/1/marathon.json#L25

wsong:
Is there any documentation for Mesos roles?  What is our scheduler supposed to do differently if a different Mesos role is specified?

kensipe:
Roles are a configuration used within Mesos for reservations.   The default is `*`, other options are arbitrary but should be configurable.   The Mesos master will only offer offers which correspond to a frameworks role.  So there isn't anything special for your framework to do differently... it is however a configuration that makes mesos more configurable and part of our requirements for certification.

there are some details under "static reservation" of this doc:  http://mesos.apache.org/documentation/reservation/

kensipe:
also @wsong there was an issue in the provided config.json.   It took some time for our CLI team to make time for it.   The file is misconfigured, however passes our schema constraints.  There is an internal jira to track this... I wanted to get the config.json update to you.  Here is a modified version of your config.json.

```
{
  "properties": {
    "memsql": {
      "additionalProperties": [],
      "properties": {
        "cpus": {
          "default": 1.0,
          "description": "CPU shares to allocate to each MemSQL scheduler instance.",
          "minimum": 0.0,
          "type": "number"
        },
        "framework": {
          "description": "Framework Scheduler specific Configuration Properties",
          "properties": {
            "authentication": {
              "description": "Framework Scheduler Authentication Configuration Properties",
              "properties": {
                "enabled": {
                  "default": false,
                  "description": "Whether framework authentication should be used",
                  "type": "boolean"
                },
                "principal": {
                  "description": "The Mesos principal used for authentication.",
                  "type": "string"
                },
                "secret": {
                  "description": "The path to the Mesos secret file containing the authentication secret.",
                  "type": "string"
                }
              },
              "required": [
                "enabled"
              ],
              "type": "object"
            }
          },
          "required": [
            "authentication"
          ],
          "type": "object"
        },
        "framework-name": {
          "default": "memsql",
          "description": "The name of the framework. Until this is configurable, please do not change this from it's default value.",
          "type": "string"
        },
        "instances": {
          "default": 1,
          "description": "Number of MemSQL scheduler instances to run.",
          "minimum": 0,
          "type": "integer"
        },
        "mem": {
          "default": 512.0,
          "description": "Memory (MB) to allocate to each MemSQL scheduler instance.",
          "minimum": 512.0,
          "type": "number"
        },
        "zk": {
          "default": "zk://master.mesos:2181/memsql",
          "description": "ZooKeeper URL for storing state. Format: host1:port1",
          "type": "string"
        }
      },
      "required": [
        "cpus",
        "framework",
        "framework-name",
        "instances",
        "mem"
      ],
      "type": "object"
    },
    "mesos": {
      "description": "Mesos specific configuration properties",
      "properties": {
        "master": {
          "default": "master.mesos:5050",
          "description": "The URL of the Mesos master.",
          "type": "string"
        }
      },
      "required": [
        "master"
      ],
      "type": "object"
    }
  },
  "required": [
    "memsql",
    "mesos"
  ],
  "type": "object"
}
```

kensipe:
please make the addition of roles to the update config.json

kensipe:
I feel like we are so close on this... sorry to be so picky on this... there is a difference between the framework-name and the framework.id.   

If you look at our chronos example, the service is started a configurable name:  
https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/C/chronos/0/marathon.json#L5

if you look at the config.json for chronos there is an ID and a frameworkname.

kensipe:
@wsong I just looked over the full docs on this... when we get the framework-name and id resolved... I think we are done.   I will give another test with that in place.   I would also like you help on verifying the service is working as expected.   It would be great to finish this today!

wsong:
Marathon uses the framework-name for its ID:

https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/M/marathon/0/marathon.json#L2

But okay, I've made them both configurable.

kensipe:
I'll ask about that!!

kensipe:
@wsong the configs and changes look good!  I was already to sign it off but then tried a standard install... it is failing.

I have a fresh dcos env... so I don't think it is the env.  Is there an issue with the service?

```
2015-07-13 20:42:34,803 [WARNING] kazoo.client: Failed connecting to Zookeeper within the connection retry policy.
2015-07-13 20:42:34,803 [INFO] kazoo.client: Zookeeper session lost, state: CLOSED
Traceback (most recent call last):
  File "/memsql_framework/bin/scheduler", line 20, in <module>
    main.run(ROOT_PATH)
  File "/memsql_framework/memsql_framework/scheduler/main.py", line 88, in run
    thread_manager, data_root, pool = setup()
  File "/memsql_framework/memsql_framework/scheduler/main.py", line 49, in setup
    root = Root(data_root_path).connect(ZOOKEEPER_URL)
  File "/memsql_framework/memsql_framework/data/root.py", line 28, in connect
    self.ZK.start()
  File "/usr/local/lib/python2.7/dist-packages/kazoo/client.py", line 537, in start
    raise self.handler.timeout_exception("Connection time-out")
kazoo.handlers.threading.TimeoutError: Connection time-out
*** /etc/my_init.d/scheduler failed with status 1

*** Killing all processes...
```

wsong:
It looks like it was because we had a zk:// prefix in our default ZOOKEEPER_URL value in config.json.  I've removed it and it seems to work.  Please take another look.

kensipe:
@wsong we are good!   If I might ask... could you pull the change of separating the id and the framework-name.   Internal discussion leads to the desire to have just one.  The framework-name.  However we still need to make the framework name as part of the marathon configuration (which was previously missing).

So to be clear:  We would like to:
1) remove the memsql.id from config
2) in marathon.json make id == memsql.framework-name
3) in marathon.json keep the newly added "MEMSQL_SCHEDULER_NAME": "{{memsql.framework-name}}"

sorry for any confusion... It was completely my fault.  We will get the chronos fixed to be the same.

kensipe:
This PR was against master... we don't use master for the universe.  It was manually merged into the version-1.x branch.  Closing this PR.

Success MemSQL is in the universe!

wsong:
Thanks Ken!

PR #162, opened 2015-06-16
Add Docker Swarm

tnachen:


elingg:
Nice, thanks guys! Are there any plans to document this as part of our release by adding the documentation here, https://docs.mesosphere.com/services/overview/? Is it certified as alpha or beta at this point? (cc @tnachen)

elingg:
@tnachen , @jimenez, @vieux, great work here! Are we going to be documenting this in our DCOS docs once this has been officially released? Is this alpha or beta quality? Has it been certified.

tnachen:
This is still alpha mostly because swarm itself is still early as well. It's not easy to hook up swarm with the cli since there isn't a clean way we can get swarm to talk over our service router. We're going to make changes directly to docker client to overcome this, but the workaround is to manually set stuff to talk to swarm at the public IP. It's usable though, so we can add some docs if people like to try it.

elingg:
Thanks for the follow up @tnachen! We can discuss with the Product Team (@keithchambers) if it makes sense to add Swarm to the docs. Meanwhile, could we make a quick change to the preInstallNotes,  "preInstallNotes": "Note that the Docker Swarm DCOS Service is alpha and there may be bugs, incomplete features, incorrect documentation or other discrepancies.", ?

vieux:
@elingg I hope the changes in docker that @tnachen just talked about will be done soon.
I would wait a bit before adding it to your docs.

elingg:
Thanks @vieux, understood.

PR #170, opened 2015-07-10
added ArangoDB framework

fceller:
New pull request for branch version-1.x


fceller:
@kensipe had to move to a new pull request to change the branch

fceller:
@kensipe had to move to a new pull request to change the branch

kensipe:
@fceller on it... I'll have feedback shortly today

kensipe:
@fceller the source looks good... I'll do some installs

repo/packages/A/arangodb/0/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "maintainer": "info@arangodb.com",
+  "description": "A distributed free and open-source database with a flexible data model for documents, graphs, and key-values. Build high performance applications using a convenient SQL-like query language or JavaScript extensions.",
+  "framework": true,
+  "tags": [ "arangodb", "NoSQL", "database", "framework" ],
+  "images": {
+    "icon-small": "https://raw.githubusercontent.com/arangodb/arangodb-dcos/master/icons/arangodb_small.png",
+    "icon-medium": "https://raw.githubusercontent.com/arangodb/arangodb-dcos/master/icons/arangodb_medium.png",
+    "icon-large": "https://raw.githubusercontent.com/arangodb/arangodb-dcos/master/icons/arangodb_large.png"
+  },
+  "preInstallNotes": "Note that the ArangoDB DCOS Service is beta and there may be bugs, incomplete features, incorrect documentation or other discrepancies.",
```
kensipe: the focus of the preinstall notes should be to inform the users what resources are necessary for this application to run.

fceller: Added resources for default setup.

kensipe:
@fceller the package install doesn't work.   Marathon just keeps cycling through installations.   There is nothing in stderr or stdout.   Have you tested this out?

fceller:
@kensipe I hope did not break anything, but we have tested it as has @joerg84, will try again

fceller:
@kensipe I did indeed break the marathon file, when replacing "/" by "." the docker-name "arangodb/arangodb-mesos" was accidentally changed to "arangodb.arangodb-mesos". I've fixed the marathon.json.

repo/packages/A/arangodb/0/config.json
```diff
@@ -0,0 +1,146 @@
...
+        "master"
+      ]
+    },
+
+    "arangodb": {
+      "description": "ArangoDB specific configuration properties",
+      "type": "object",
+      "properties": {
+
+        "id": {
```
kensipe: this probably seems crazy :) please switch the property id to framework-name.   or add framework-name.   we are standardizing on using just 1 of these properties `framework-name` hence the request to change the name.  Note that there is special magic that ties the health check display on the dcos ui to the framework if 2 things are true:  1) the framework has a property of framework: true (arangodb already does) and 2) there is a property of framework-name with a default value.

Also note... a change of id here requires a change of id under the required section plus changes to all references of id in the marathon.json

kensipe:
@fceller please note above need for `framework-name` property change.  Also you will need to change the http service in the scheduler.  the assets are absolute path... which I'm sure works feel in testing, however the service is proxied and the assets are lost so I can't see the ui admin screen.

The link from the dcos ui will result in:  http://ken-memsq-elasticl-oncspqsl4gbw-1318642894.us-west-2.elb.amazonaws.com/service/arangodb/
from that location you have 4 assets with issues.. it is all 4 scripts after the footer, starting with `js/generated/libs.js`

repo/packages/A/arangodb/0/config.json
fceller: @kensipe I've changed "id" to "framework-name" and removed the "id" also from the marathon.json

kensipe:
@fceller we are super close on this... I need to follow up with a couple of things... I need you to make 1 change to the CLI to use the admin router.  I found an example:

https://github.com/mesosphere/dcos-kafka/blob/master/dcos_kafka/cli.py#L34-L35

here is the issue... they are hardcoding it...  am I'm going to ask you to do the same for right now. (of course to arangodb).   The problem is if anyone overrides the framework-name the CLI will break.   I don't know internally how we are expecting to handle this yet.

kensipe:
@fceller if we can get that CLI fix in... I would like to merge this in to the universe tomorrow.

fceller:
@kensipe I have added the "rstrip".

But you lost me with the hardcoded stuff. Currently I can do the following:

```
# cat name1.json 
{ "arangodb" : { "framework-name" : "frank" } }

# cat name2.json 
{ "arangodb" : { "framework-name" : "ken" } }

# dcos package install arangodb --options=name1.json 
# dcos package install arangodb --options=name1.json 
```

to install two ArangoDB clusters, switch to the corresponding FE of either "frank" or "ken". And then use uninstall one, using

```
# dcos arangodb uninstall --internal --app-id ken 
# dcos package uninstall --app-id=ken arangodb
```

If I hardcoded the "arangodb" name, none of this will work any longer.

kensipe:
@fceller I would rather you not hardcode :) 

I'm looking for your CLI sub commands to hit the service endpoint.  So if when we are there let me know.   I provided a link to the kafka example earlier.  (they are hardcoding to the kafka name... if you don't have to... that's awesome).   when this is done, I think we are ready to merge .

fceller:
Hi Ken, can you try the newest version of the ArangoDB CLI? It should now generate:

root@f66aecf88a54:~/dcos# dcos arangodb webui
http://104.46.40.104/services/arangodb

fceller:
@kensipe I have just seen that I made a type. It "service" not "services", correct?

fceller:
@kensipe it is now "service"

PR #173, opened 2015-07-20
Kubernetes v1.0-pre

tsenart:
This change set upgrade the Kubernetes package to that latest upstream v1.0 code.

Two remarks regarding the universe structure:
* Having a compressed index in Git makes conflicts inevitable. We need a better solution.
* There's too much duplication between package versions. We need to find a way to reuse more stuff.

Ref: https://github.com/mesosphere/kubernetes-mesos/issues/371


karlkfi:
Is there an easy way to point a DCOS deploy at this PR to test the updated k8s?

tsenart:
@karlkfi: I think @sttts answered that question on Slack. You can checkout the right branch locally and point your `~/.dcos/dcos.toml` to it with the `file://` prefix. Don't forget to run `dcos package update` after that.

PR #174, opened 2015-07-20
fix memsql config schema

tamarrow:
Don't allow additionalProperties. Must be boolean or object to be a
valid jsonschema.


kensipe:
lgtm... tested on dcos... all is well!

PR #179, opened 2015-07-21
updated kafka-mesos version to kafka-mesos-0.9.1.1.jar

dmitrypekar:
Hey, I've update kafka-mesos version on S3 to 0.9.1.1. 
Please merge.


BenWhitehead:
@dmitrypekar @tnachen In the future when a new version of a package is to be released there should be a new revision of the package rather than changing the existing revision.

/cc @jsancio

PR #180, opened 2015-07-21
Chronos network bridge.

enxebre:


BenWhitehead:
@enxebre Can you explain the motivation for this change?

enxebre:
Hi there!
First reason is that current approach did not work for me as chronos was trying to register continuously as it was been seen as localhost by mesos master.

Second reason is avoiding embracing network isolation by using "network": "BRIDGE"
https://github.com/docker/docker/issues/6401
https://github.com/docker/docker/issues/14767

Thanks!

BenWhitehead:
You're right in that we want to try and avoid the scenario mentioned in those two docker tickets, however the current implementation of mesos requires that host networking be used when running a framework inside a docker container. This is due to the fact that libmesos opens a random socket in order to receive messages from the master.

Please double check your logs and ensure that when you have bridge networking enable for chronos you see a log line similar to the following in `stderr`:
```
sched.cpp:448] Framework registered with 20150723-182640-3540320266-5050-1210-0019
```
If you don't see this line, chronos didn't actually fully register as a framework with mesos and won't be able to receive resource offers.

enxebre:
yup, is definitely working for me when using bridge networking, you can see it in action at the minute here http://52.18.143.92:8080/ and http://52.18.143.92:5050  and http://52.17.148.221:31787/#jobs/test

But thanks to your comment just realised that might be due to the fact that in our scenario all containers are bind to a weave network (http://docs.weave.works/weave/latest_release/how-it-works.html). Does it make sense for you?

BenWhitehead:
Ah yes, weave definitely changes the behavior.  Thanks for the additional details.  When running on DCOS without any extra networking layer the bridge networking option will not work, so I suspect this PR won't be merged since we won't be able to guarantee that weave will be running for all DCOS clusters.

I'm going to defer to @jsancio for the final decision though. Thanks for helping me understand.

enxebre:
Sure, that make sense. Sorry haven't reallised about weave was coming the game before.
Thank you!!

enxebre:
How about making network parameter overridable from the config.json file? Have you considered that posibility?

PR #181, opened 2015-07-21
Jd: update to use version 0.9.1 kafka-mesos jar

john-dodson:
Update to use new kafka-mesos jar version 0.9.1 that includes bug fixes necessary for running kafka-mesos on instances other than m3.xlarge


BenWhitehead:
I this is fixed by #179 which has already been merged.

PR #182, opened 2015-07-21
Update HDFS CLI Version

elingg:


BenWhitehead:
@elingg We should get a pip build setup for this as we're discouraging the use of git URLs for installs.

Also releasing an updated version of a package a new package revision should be created rather than changing the existing revision.

elingg:
@BenWhitehead, 

Yes, we need to set up build automation for the CLI. This is a known issue that needs to be addressed.

Regarding package revision, makes sense. We probably want to sync our releases of the framework and the CLI in that case. Fortunately, we are also checking in version 0.1.2 of the framework right now along with updating the version of the package. This is a bit of a tricky issue, though, because we may have different versions of the framework and the CLI.

PR #183, opened 2015-07-21
Package hadoop and java with the HDFS framework

gabrielhartmann:
Frameworks in the continuous/testing branch no longer have access to the
host's java and hadoop.  It must be packaged with the framework.

### DIFF
--- a/repo/packages/H/hdfs/0/config.json
+++ b/repo/packages/H/hdfs/0/config.json
@@ -11,7 +11,8 @@
           "type": "string"
         },
         "default": [
-          "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz"
+          "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz",
+          "https://downloads.mesosphere.io/java/jre-7u76-linux-x64.tar.gz"
         ]
       },
       "custom-config": {
@@ -22,11 +23,29 @@
         "description": "The name of the framework. Until this is configurable, please do not change this from it's default value.",
         "type": "string",
         "default": "hdfs"
+      },
+      "framework-version": {
+        "description": "Framework version",
+        "type": "string",
+        "default": "hdfs-mesos-0.1.1"
+      },
+      "jre-version": {
+        "description": "JRE version",
+        "type": "string",
+        "default": "jre1.7.0_76"
+      },
+      "ld-library-path": {
+        "desription": "Library path",
+        "type": "string",
+        "default": "/opt/mesosphere/lib"
       }
      },
      "required": [
       "scheduler-uris",
-      "framework-name"
+      "framework-name",
+      "ld-library-path",
+      "jre-version",
+      "framework-version"
      ]
     } 
   }
--- a/repo/packages/H/hdfs/0/marathon.json
+++ b/repo/packages/H/hdfs/0/marathon.json
@@ -4,12 +4,15 @@
   "mem": 512,
   "disk": 100,
   "instances": 1,
-  "cmd": "{{#hdfs.custom-config}}echo {{hdfs.custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs.custom-config}}cd hdfs-mesos* && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
+  "env": {
+    "LD_LIBRARY_PATH": "{{hdfs.ld-library-path}}" 
+  },
+  "cmd": "export JAVA_HOME=$MESOS_DIRECTORY/{{hdfs.jre-version}} && {{#hdfs.custom-config}}echo {{hdfs.custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs.custom-config}}cd {{hdfs.framework-version}} && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
   "uris": {{{hdfs.scheduler-uris}}},
   "healthChecks": [
     {
       "protocol": "COMMAND",
-      "command": { "value": "hadoop fs -ls hdfs://hdfs/ && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -mkdir hdfs://hdfs/hdfs-framework-healthcheck && mkdir hdfs-framework-healthcheck && echo \"this is a test\" > hdfs-framework-healthcheck/test1.txt && hadoop fs -put hdfs-framework-healthcheck/test1.txt hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -get hdfs://hdfs/hdfs-framework-healthcheck/test1.txt hdfs-framework-healthcheck/test2.txt && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck" },
+      "command": { "value": "export PATH=$MESOS_DIRECTORY/{{hdfs.framework-version}}/bin:$PATH && export JAVA_HOME=$MESOS_DIRECTORY/{{hdfs.jre-version}} && hadoop fs -ls hdfs://hdfs/ && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -mkdir hdfs://hdfs/hdfs-framework-healthcheck && mkdir hdfs-framework-healthcheck && echo \"this is a test\" > hdfs-framework-healthcheck/test1.txt && hadoop fs -put hdfs-framework-healthcheck/test1.txt hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -get hdfs://hdfs/hdfs-framework-healthcheck/test1.txt hdfs-framework-healthcheck/test2.txt && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck" },
       "gracePeriodSeconds": 300,
       "intervalSeconds": 60,
       "timeoutSeconds": 20,


repo/packages/H/hdfs/0/config.json
```diff
@@ -11,7 +11,8 @@
           "type": "string"
         },
         "default": [
-          "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz"
+          "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz",
```
gabrielhartmann: I'm not sure how the upgrade for this works.  At what point will this hdfs-mesos package be the correct one from [the other PR](https://github.com/mesosphere/hdfs/pull/157)?

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -4,12 +4,12 @@
   "mem": 512,
   "disk": 100,
   "instances": 1,
-  "cmd": "{{#hdfs.custom-config}}echo {{hdfs.custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs.custom-config}}cd hdfs-mesos* && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
```
elingg: We should avoid hard coding LD_LIBRARY_PATH and use configuration values similar to what is done in kafka, i.e. {{hdfs.java-home}}

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -4,12 +4,12 @@
...
+  "cmd": "export LD_LIBRARY_PATH=/opt/mesosphere/lib && export JAVA_HOME=$MESOS_DIRECTORY/jre1.7.0_76 && env && {{#hdfs.custom-config}}echo {{hdfs.custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs.custom-config}}cd hdfs-mesos-* && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
   "uris": {{{hdfs.scheduler-uris}}},
```
elingg: Just a question do we need ```&& env``` here? It seems to be just printing the environment so I don't think we need it.

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -4,12 +4,12 @@
...
+      "command": { "value": "export PATH=$MESOS_DIRECTORY/hdfs-mesos-0.1.1/bin:$PATH && export JAVA_HOME=$MESOS_DIRECTORY/jre1.7.0_76 && hadoop fs -ls hdfs://hdfs/ && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -mkdir hdfs://hdfs/hdfs-framework-healthcheck && mkdir hdfs-framework-healthcheck && echo \"this is a test\" > hdfs-framework-healthcheck/test1.txt && hadoop fs -put hdfs-framework-healthcheck/test1.txt hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -get hdfs://hdfs/hdfs-framework-healthcheck/test1.txt hdfs-framework-healthcheck/test2.txt && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck" },
       "gracePeriodSeconds": 300,
```
elingg: Same comment as above. Let's avoid the hard coding and use configuration values similar to kafka.

repo/packages/H/hdfs/0/config.json
```diff
@@ -11,7 +11,8 @@
...
+          "https://downloads.mesosphere.io/java/jre-7u76-linux-x64.tar.gz"
         ]
```
elingg: Let's avoid the hard coding and use configuration values similar to kafka.

elingg:
Great, thanks @gabrielhartmann! Some comments above

repo/packages/H/hdfs/0/config.json
elingg: Great question @gabrielhartmann. You can ping me, and I will walk you through the upgrade process once the comments are addressed. You can merge this PR in Universe after the other PR is merged and we release a new version of HDFS. First step will be getting a team city account from Ben W.

repo/packages/H/hdfs/0/marathon.json
elingg: Also, instead of export, we should definitely use env variables. See https://mesosphere.github.io/marathon/docs/rest-api.html to see how a user was using LD_LIBRARY_PATH as an env variable with their cmd

repo/packages/H/hdfs/0/config.json
gabrielhartmann: Are you referring to "https://downloads.mesosphere.io/java/jre-7u76-linux-x64.tar.gz" when you are speaking about not hard coding here?  This change is in the config file.

repo/packages/H/hdfs/0/marathon.json
gabrielhartmann: I'm moving the setting of LD_LIBRARY_PATH to the "env" section.

> Just a question do we need && env here?
It's not strictly necessary, but nice to have for debugging purposes, so one can see what environment the scheduler sees.

repo/packages/H/hdfs/0/config.json
elingg: Sorry, I think I read too quickly here. We should be using this config value in Marathon.json and not hard coding there. This one looks good.

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -4,12 +4,15 @@
...
+      "command": { "value": "export PATH=$MESOS_DIRECTORY/{{hdfs.framework_version}}/bin:$PATH && export JAVA_HOME=$MESOS_DIRECTORY/{{jre_version}} && hadoop fs -ls hdfs://hdfs/ && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -mkdir hdfs://hdfs/hdfs-framework-healthcheck && mkdir hdfs-framework-healthcheck && echo \"this is a test\" > hdfs-framework-healthcheck/test1.txt && hadoop fs -put hdfs-framework-healthcheck/test1.txt hdfs://hdfs/hdfs-framework-healthcheck && hadoop fs -get hdfs://hdfs/hdfs-framework-healthcheck/test1.txt hdfs-framework-healthcheck/test2.txt && rm -rf hdfs-framework-healthcheck && hadoop fs -rm -r -f hdfs://hdfs/hdfs-framework-healthcheck" },
```
gabrielhartmann: Typo.
export JAVA_HOME=$MESOS_DIRECTORY/{{jre_version}} should be
export JAVA_HOME=$MESOS_DIRECTORY/{{hdfs.jre_version}}

repo/packages/H/hdfs/0/marathon.json
gabrielhartmann: Resolved.

repo/packages/H/hdfs/0/marathon.json
elingg: Let's remove the export statements and then put JAVA_HOME, etc in the env variables section to make this cleaner.

repo/packages/H/hdfs/0/config.json
```diff
@@ -26,7 +42,10 @@
      },
      "required": [
       "scheduler-uris",
-      "framework-name"
+      "framework-name",
```
elingg: We need to use dashes, not underscores here. That is the consistent format used in the universe packaging.

repo/packages/H/hdfs/0/marathon.json
gabrielhartmann: There is no environment section for health checks. Also you can't move
anything into the cmd environment which references another environment
variable.  Now we have three different ways of expressing environment.
Just documenting this.  We spoke about this face-to-face.

elingg: Yes, great points, we discussed. 

For health checks that needs to be ultimately moved to the scheduler, so that's a TODO for clean up.

I would prefer that env could do the parsing the cmd does. Unfortunately, our JSON implementation doesn't allow this currently and it would be nice if this was improved in the CLI. Meanwhile, it makes sense to pull out whatever we can into the command env so we don't have one giant command.

repo/packages/H/hdfs/0/config.json
```diff
@@ -11,9 +11,25 @@
           "type": "string"
         },
         "default": [
-          "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz"
+          "https://s3.amazonaws.com/downloads.mesosphere.io/hdfs/artifacts/dcos/0.1.1/hdfs-mesos-0.1.1.tgz",
+          "https://downloads.mesosphere.io/java/jre-7u76-linux-x64.tar.gz"
         ]
       },
+      "ld-library-path": {
```
elingg: These values should be in alphabetical order. They seem to be in reverse alphabetical order.

repo/packages/H/hdfs/0/marathon.json
```diff
@@ -4,12 +4,15 @@
   "mem": 512,
   "disk": 100,
   "instances": 1,
-  "cmd": "{{#hdfs.custom-config}}echo {{hdfs.custom-config}} | base64 --decode > mesos-site.xml && {{/hdfs.custom-config}}cd hdfs-mesos* && if [ -e '../mesos-site.xml' ]; then cp -f ../mesos-site.xml ./etc/hadoop/; fi && ./bin/hdfs-mesos",
+  "env": {
+    "LD_LIBRARY_PATH": "{{hdfs.ld-library-path}}" 
+  },
```
elingg: Print env is useful for debugging, but in the version we check in, we assume debugging is not required. I think we should be consistent and not display env here since we do that do this as a standard elsewhere. Also, it will keep the command a little simpler.

elingg:
Ok two final comments to address, @gabrielhartmann  but this is looking great! Any objections @kensipe ? If not, we can upgrade the version of HDFS and then release this.

elingg:
Looks like you accidentally duplicated the values in config.json, but other than that, I think we are good

BenWhitehead:
@elingg @gabrielhartmann in the future, when a new version of a package is being released there should be a new package revision in the repo rather than changing the existing revision.

gabrielhartmann:
I screwed this up.  I misunderstood the order of operations on releasing a
new package.

elingg:
Yes, makes sense, he was new to the process, which is not 100% documented. We were supposed to merge only after updating the versions and index file.

PR #184, opened 2015-07-22
kubernetes: v1.0.1-v0.6.0-alpha

tsenart:


tsenart:
We smoke tested the guestbook but would appreciate someone else doing it too before merging this in.

karlkfi:
Testing:
```
sed -i.bkp -e "s@sources = \[.*\]@sources = \[\"https://github.com/mesosphere/universe/archive/kubernetes-v1.0.1-v0.6.0-alpha.zip\"\]@" ~/.dcos/dcos.toml
dcos package update
```

Remove backup config:
```
rm ~/.dcos/dcos.toml.bak
```

karlkfi:
IS there a way to do this and update the dcos docs at the same time?

tsenart:
We're working on updating [the docs](https://github.com/mesosphere/mesosphere-docs/blob/master/services/kubernetes.md) tomorrow. Although that will be a separate PR on a separate repo, we'll try to sync their release.

karlkfi:
I just rebuilt the guestbook-go with relative links. So if you change the gustbook-controller to use the `mesosphere/kubernetes-guestbook-go:latest` image it will work through the dcos proxy: `http://<host>/service/kubernetes/api/v1/proxy/namespaces/default/services/guestbook/`

repo/packages/K/kubernetes/1/config.json
```diff
@@ -0,0 +1,195 @@
...
+          "description": "Docker image used to launch the Kubernetes framework on DCOS.",
+          "type": "string"
+        },
+        "dns-suffix": {
+          "default": ".marathon.mesos",
+          "description": "This value is appended to the app-id value to form the canonical DNS name for the Kubernetes components.",
+          "type": "string",
+          "pattern": "^(?:\\.[a-z][a-z0-9]*?(?:-[a-z0-9]+)*)+$"
+        },
+        "enable-dns": {
```
jdef: it looks like the shell scripting is expecting a bool now: `true` instead of `1`. so this needs to change

repo/packages/K/kubernetes/1/config.json
```diff
@@ -0,0 +1,195 @@
...
+          "pattern": "^1?$"
+        },
+        "framework-name": {
+          "default": "kubernetes",
+          "description": "Human-readable name to use when registering with the Mesos master.",
+          "type": "string",
+          "pattern": "^(?:[a-z][a-z0-9]*?(?:-[a-z0-9]+)*)$"
+        },
+        "kube-dns-domain": {
+          "default": "kubernetes.local",
```
jdef: looks like the k8s default is now `cluster.local`; we should probably strive for consistency here

repo/packages/K/kubernetes/1/config.json
```diff
@@ -0,0 +1,195 @@
...
+          "type": "string",
+          "pattern": "^(?:(?:[0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}(?:[0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(?:\\/(?:[0-9]|[1-2][0-9]|3[0-2]))$"
+        },
+        "scheduler-service-ip": {
+          "default": "10.10.10.9",
+          "description": "Kubernetes-Mesos scheduler service portal IP address, should fall in service-cluster-ip-range network range.",
+          "type": "string",
+          "pattern": "^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
+        },
+        "disable-etcd-server": {
```
jdef: if `enable-dns` becomes a bool then this should as well, for consistency. will require bootstrap.sh updates.

tsenart: Makes sense.

repo/packages/K/kubernetes/1/config.json
tsenart: Good catch.

repo/packages/K/kubernetes/1/config.json
tsenart: Sounds good.

tsenart:
@jdef: PTAL

jdef:
lgtm

sttts:
This one can be closed. #191 is the follow-up.

PR #189, opened 2015-07-30
releasing hdfs 0.1.3

kensipe:
@elingg a review would be nice!


elingg:
Looks perfect! As long as this is tested, merge away

PR #190, opened 2015-07-30
Add LD_LIBRARY_PATH to environment

gabrielhartmann:
This change allows Cassandra can be deployed on DCOS Test/Continuous AWS clusters.  It still deploys  fine on Stable clusters too.  When a new release of Cassandra comes out the URI will have to be updated here, but so far this is a non-breaking change.

### DIFF
--- a/repo/packages/C/cassandra/0/config.json
+++ b/repo/packages/C/cassandra/0/config.json
@@ -71,6 +71,11 @@
           "type": "string",
           "default": "dcos"
         },
+        "ld-library-path": {
+          "desription": "Library path",
+          "type": "string",
+          "default": "/opt/mesosphere/lib"
+        },
         "zk": {
           "description": "ZooKeeper URL for storing state. Format: zk://host1:port1,host2:port2,.../path (can have nested directories)",
           "type": "string"
@@ -163,5 +168,4 @@
     "mesos",
     "cassandra"
   ]
-
 }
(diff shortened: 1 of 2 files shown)


repo/packages/C/cassandra/0/config.json
```diff
@@ -71,6 +71,11 @@
           "type": "string",
           "default": "dcos"
         },
+        "ld-library-path": {
```
elingg: nit: description should say LD Library path

elingg:
one tiny nit, but LGTM @gabrielhartmann!

elingg:
Also, could you please create a new package definition for Cassandra? After chatting with Ben and Jose, going forward, for every modification we need to create a new package definition. This would be Version 1.

BenWhitehead:
We should not be providing a parameter for the user to change the LD_LIBRARY_PATH themselves when it is ran on dcos.  DCOS has a controlled environment and stuff will always be in those locations.

The literal value should be put into the marathon.json.

BenWhitehead:
Also, as @elingg mentioned the intent of universe is to be append only. This will require a new package revision.

elingg:
@BenWhitehead, why not allow the user to override it? I think it is reasonable, especially in the future we may package the libmesos library along with our framework.

elingg:
We discussed and we'll hard code it in marathon.json. We need to figure out what our standard process is around frameworks packaging in Universe. Automative tests will be a big help here.

PR #192, opened 2015-08-05
Update spark to 1.4.1 and client to 0.2.0

tnachen:
This includes upgrading spark to 1.4.1, and also let the client use spark 1.4.1 image as well.
There are also fixes that are applied to client and new features such as tailing driver log.


kensipe:
@tnachen starting review to merge

repo/packages/S/spark/1/marathon.json
```diff
@@ -0,0 +1,36 @@
+{
+  "id": "spark",
```
kensipe: we are looking for id, cpus, mem, instances to all have config.json properties so they can be modified at installation

tnachen: Oh? even id? ok let me make this change

kensipe:
@tnachen I rebased to the latest and I will update with the needed changes unless you are in the middle of it...

PR #204, opened 2015-08-17
Package for Marathon 0.9.2

kolloch:
and added tons of missing options.

See https://github.com/mesosphere/marathon/issues/2015 for the
Marathon issue tracking this.


kolloch:
Hi @kensipe, @aquamatthias recommended you for the best review service ;)

kolloch:
This might be helpful for testing:

https://mesosphere.atlassian.net/wiki/display/MSPHERE/Universe+Package

If there is already documentation like that, let me know ;)

kensipe:
lol... you got it @kolloch ... for you... only the best!

kensipe:
LGTM... let me give it a test run... then I'll merge

kensipe:
glad I tested... there are no health checks on this... don't you want to be healthy? @kolloch

kolloch:
That's a valid concern ;) There wasn't any health check in prior configurations. I added one.

Note that we do not have a good "health" endpoint. (yes, that's a shame)

Unfortunately, I already trashed my DCOS cluster and will not be able to test that today anymore. If, by chance, you can quickly check that, I would be very thankful. We could avoid another time zone round trip.

If not, I will test it tomorrow. Thanks @kensipe !

kensipe:
SHHHHH testing in progress

kensipe:
the json is bad... looking into it

kolloch:
thanks, @kensipe. Strange, I quickly tested it locally and at least I did not run into validation errors.

kolloch:
[My local setup does not allow Marathon to connect to the master, though. Hence the note about not testing it.]

kensipe:
i just double checked... it looks like an issue on my end... checking again.

kensipe:
health checks fail and cause a failed deployment... checking deeper

repo/packages/M/marathon/3/config.json
```diff
@@ -0,0 +1,287 @@
...
+          "default": [],
+          "description": "List of URIs that will be download and made available in the current working directory of Marathon. For example this can be used to download a Java keystore file for SSL configuration.",
+          "items": {
+            "pattern": "^[\\s]+",
+            "type": "string"
+          },
+          "type": "array"
+        },
+        "zk": {
+          "default": "zk://master.mesos:2181/universe",
```
kensipe: I hate to be picky... but "universe" is a bad name.   This is another instance of a marathon that is NOT DCOS.  Also... the zk should probably be dynamic in if someone wanted 1 or 2 of these installed they could give me different framework-names and they would have their own namespace in zk.

regardless universe is not a good name for this IMO

kensipe:
ok... I'm an idiot :)  the issue was clearing out zk... health checks are fine.   I'll merge this... however I really do think we should rename the zk namespace... anything but universe.   please submit another PR for that when you get a chance.

repo/packages/M/marathon/3/config.json
kolloch: I completely agree! I just copied it from the other config file.

It would be even better to base the ZK path on the app-id or something like that (if it is possible).

Unfortunately, I think its even worse to change the default now. If someone does upgrade from a prior package definition, the Marathon will not contain the old definitions anymore.

Do you agree or can you think of a way to mitigate that?

kolloch:
Thanks for the check and the merge!

kolloch:
Oh, and by the way: I do not object to being picky in a pull request in any way! Please, always be picky.

PR #206, opened 2015-08-18
Package Datadog Agent for DCOS to monitor mesos

alq666:
[Datadog](http://datadog.com) can monitor mesos and marathon natively (this version only support mesos). It uses an agent that runs as a docker container and queries the stats endpoints.

Here is what it looks like when running on DCOS:

screenshot 2015-08-18 16 54 14


alq666:
The conflict stems from `index.json.gz` being checked in, by the way.

repo/packages/D/datadog/0/marathon.json
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    "key": "name", "value": "dd-agent"
+                },
+                {
+                    "key": "env", "value": "API_KEY={{datadog.api_key}}"
+                }
+            ],
+            "network": "BRIDGE",
+            "portMappings": [
+                { "hostPort": 8125, "containerPort": 8125, "protocol": "udp" },
```
ConnorDoyle: Could you please make the `hostPort` configurable for both of these?  Maybe you could default to the canonical values.

repo/packages/D/datadog/0/marathon.json
```diff
@@ -0,0 +1,59 @@
+{
+    "id": "datadog-agent",
```
kensipe: we need this to be configurable... it is commonly framework-name, however this doesn't seem to be a framework.  We still need this to be a configurable {{data-dog.app-id}}

repo/packages/D/datadog/0/marathon.json
alq666: Sure thing.

repo/packages/D/datadog/0/marathon.json
alq666: Will do.

repo/packages/D/datadog/0/marathon.json
```diff
@@ -0,0 +1,62 @@
+{
+    "id": "{{datadog.app_id}}",
```
alq666: @kensipe does that work as a configurable?

kensipe: perfect! 
got busy with mesoscon etc... but I'm on this now.

kensipe:
@alq666 I'm checking this out... the installation fails due to licensing with the following message:

```
Registered docker executor on ip-10-0-0-206.us-west-2.compute.internal
Starting task datadog-agent.e2b9de24-4c04-11e5-87d3-521641bcf454
You must set API_KEY environment variable to run the Datadog Agent container
```

1. can you help me out here
2. is there a place where this is documented?

kensipe:
@alq666 
is it expected that a user must create a json file to over these values at the time of installation?

```
Pass your API key as a package option (datadog.api_key).
Also pass the number of Mesos slaves as an option (datadog.instances).
```

alq666:
@kensipe it is so I'm curious if there's a better way to do this.

The API key is an obvious one. The number of instances is until we can run as an admin task (and it gets deployed on all the agent nodes).

kensipe:
Tested... LGTM.  I'm running details past our internal team.

kensipe:
@alq666 here is what I would like to do... will you remove the default value for api_key in config.json?  By removing it,  the DCOS CLI will see that it is required but that it is missing... this will fail the installation until a key is put in a json file and included with the `--options` flag.     This will be a better experience IMO.   After that... I'm inclined to merge this.  nice work!

kensipe:
a bit usually... but I've tested the change I requested... so I will merge manual with the change and add this in the universe.

alq666:
@kensipe thank you! :+1:

PR #207, opened 2015-08-20
Update chronos marathon.json with updated image

wallies:
Need this to work with latest mesos


kensipe:
a version update like this would require a dcos version update... so this would increment chronos dcos version from 0 to 1.. and add an entry in the index.

Is that something you can add to this PR or should I create another and close this one?

wallies:
i think i have done this correctly. Is there something I am missing?

kensipe:
first glance it looks correct.  I'll comb thru and test now.

kensipe:
and thx!

kensipe:
LGTM and tested!  Thanks for the help!

PR #213, opened 2015-09-01
Repoint the Datadog Agent docker image to the official Datadog Docker Hub image

alq666:
Rather than a build from my user account.

No functionality change: Both are building from the same branch: https://github.com/DataDog/docker-dd-agent/tree/alq666/mesos.


alq666:
@kensipe looks like we're not in the universe anymore (based on your commit message). Let me know once there's a good place for our stuff to live.

PR #214, opened 2015-09-04
Minor update of the ArangoDB subcommand.

neunhoef:
I bumped the version to 0.2.1.
This is basically only a cleanup of the configuration options. In particular, I have added configuration options for
  - authentication with the mesos master
  - two switches for asynchronous replicas


neunhoef:
I have no access to the results of the TeamCity build tests. Could somebody please let me know what is wrong, since it seems to be triggered by my changes?

repo/packages/A/arangodb/1/command.json
```diff
@@ -0,0 +1,5 @@
+{
+  "pip": [
+    "git+https://github.com/arangodb/arangodb-dcos"
```
kensipe: we have previously allowed this.. however internally we discussed this and we would prefer to see the link for sub command be for a specific version and not off master.   we don't want changes to the master to be automatically added in without a versioning of the dcos service.   the idea is to have an immutable version.

neunhoef:
OK, no problem. What is the syntax then for a branch or tag say? Is there any documentation on this?
According to the messages, this does not seem to be the cause of the problem.

neunhoef:
From my perspective this looks now good.

kensipe:
lgtm... merging

PR #220, opened 2015-09-12
fix mesos auth in cassandra marathon templates

### LINKED ISSUE #219: incorrect marathon templates for cassandra
In the cassandra package, the keys (defined in config.json) for specifying mesos authentication principal and secret are "cassandra.framework.authentication.principal" and "cassandra.framework.authentication.secret". However, in the marathon.json template, the mustache sections are keyed with just "framework.authentication.principal" and "framework.authentication.secret":

https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/C/cassandra/1/marathon.json#L67
https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/C/cassandra/1/marathon.json#L70

So, if you turn on authentication using an options file when installing cassandra, the authentication information is never set and the cassandra package will not install correctly.

ryane:
fixes #219


BenWhitehead:
Thanks @ryane!

PR #221, opened 2015-09-14
Add prometheus

discordianfish:
This adds the prometheus demo from the dreamathon. I like to deploy that to our new cluster.
Let me know if we can merge it or not, given it's not officially supported.


kensipe:
@discordianfish we are moving to a model to add services into the multiverse first... especially for new or experimental services.   could you send a PR against https://github.com/mesosphere/multiverse/ on the master?

discordianfish:
Yes, sounds good!

Issue #222, opened 2015-09-15
Elasticsearch fails to install with Unknown Option -zk


jgarcia-mesosphere:
When using `dcos package install elasticsearch`, it fails with the error 
```/bin/sh -c exit `docker wait mesos-fc4657c0-2589-4d93-8f0f-67ed79bbb945` 
com.beust.jcommander.ParameterException: Unknown option: -zk```

enxebre:
same here, think that the marathon file needs to be updated according to the java process options

jsancio:
ElasticSearch is not a package in this repository any more.

PR #223, opened 2015-09-15
Fixing Unknown Option -zk

jgarcia-mesosphere:
More info at issue #222


jgarcia-mesosphere:
This actually was reverted via force-push

PR #225, opened 2015-09-18
Marathon has a bug.  Health checks cannot timeout.

gabrielhartmann:
https://github.com/mesosphere/marathon/issues/2078


elingg:
:+1:

kensipe:
+1

PR #231, opened 2015-09-22
making cassandra framework resources configurable

kensipe:


kensipe:
@elingg looking for review on this...  I am looking to test still... back in a few with the results.

elingg:
LGTM assuming this is tested

elingg:
Possibly this could be an append only change and we should create a new folder for this, however, I feel it is quite small. I could go either way.

kensipe:
tested good!!

repo/packages/C/cassandra/1/config.json
```diff
@@ -31,6 +31,23 @@
...
+              "description": "CPU shares to allocate to each Cassandra framework instance.",
+              "type": "number"
+            },
+            "mem": {
+              "default": 512.0,
+              "description": "Memory (MB) to allocate to each Cassandra framework instance.",
+              "minimum": 512.0,
+              "type": "number"
+            },
+            "instances": {
```
BenWhitehead: This should not be configurable.  We only ever want one instance of the scheduler running at a time.  The number of nodes in the cluster is configured by another parameter.

repo/packages/C/cassandra/1/marathon.json
```diff
@@ -1,9 +1,9 @@
 {
   "id": "/cassandra/{{cassandra.cluster-name}}",
   "cmd": "$(pwd)/jre*/bin/java $JAVA_OPTS -classpath cassandra-mesos-framework.jar io.mesosphere.mesos.frameworks.cassandra.framework.Main",
-  "instances": 1,
-  "cpus": 0.5,
-  "mem": 512,
+  "instances": {{cassandra.framework.instances}},
```
BenWhitehead: See not above in config.json

elingg:
Oh I missed this. yes we want only once instance of the scheduler at this point cc @kensipe. Making mem and cpus should be configurable though.

PR #232, opened 2015-09-22
Remove binary index

kensipe:
the binary file is a common source of PR collisions.  It isn't used and this PR removes it.


kensipe:
@jsancio and @elingg looking for a review.  It is important to note that I have done manual tests verifying that this PR is good and I've removed the binary from the cli-tests branch and ran full tests on the cli which passed.

elingg:
@Kensipe Removing the compressed index file has no effect? In that case, I do think it makes sense to remove for a better user experience. Are any useful checks in the build script getting removed by removing this file?

kensipe:
all checks still exist.. the only script change was the removal of code to build the gz file.

elingg:
LGTM

kensipe:
merging... this has been aggressively tested and had several eyeballs on it.

PR #238, opened 2015-09-30
Force a timeout of the healthcheck in 45s

gabrielhartmann:
Until https://issues.apache.org/jira/browse/MESOS-3479 is resolved we
must never timeout.  There are cases where we still exceed the 60s
timeout

I got lucky testing this.  I ran into the scenario which usually has a health check exceed it's 60s timeout, but the timeout enforced here (Ctrl+f timeout 45s) caused the health-check to fail on time.  The next health-check succeeded as usual, but this time we converged to a healthy state.  Defect mitigated!


elingg:
LGTM

PR #239, opened 2015-10-01
Update preinstallNotes for clarity

elingg:


elingg:
@mgummelt or @gabrielhartmann, please review

repo/packages/H/hdfs/1/package.json
```diff
@@ -17,7 +17,7 @@
       "url": "https://github.com/mesosphere/hdfs/blob/master/LICENSE"
     }
   ],
-  "preInstallNotes": "In order for HDFS to start successfully, all resources must be available in the cluster, including ports, CPU shares, RAM and disk.\nWe recommend a minimum of five nodes, each with at least 2 CPU shares and 8GB of RAM available for use by the HDFS Service.\nNote that the service is alpha and there may be bugs, including possible data loss, incomplete features, incorrect documentation or other discrepancies.",
+  "preInstallNotes": "In order for HDFS to start successfully, we recommend a minimum of five nodes, each with at least 2 CPU shares and 8GB of RAM available for use by the HDFS Service.\nNote that the service is alpha and there may be bugs, including possible data loss, incomplete features, incorrect documentation or other discrepancies.",
```
mgummelt: "five nodes" isn't accurate, actually.  The nodes must be private.  But we don't expose this well, so I'm not sure if we should chagne it.

mgummelt:
LGTM

repo/packages/H/hdfs/1/package.json
elingg: Yep and actually we need a minimum for 4 nodes for our set up but we say 5 so we can have 2 datanodes for replication

PR #242, opened 2015-10-06
spark 1.5.0

mgummelt:
@elingg @tnachen please review

This has been manually tested with the dcos-tests suite as well as a TeraSort test.

### DIFF
--- a/repo/packages/S/spark/2/config.json
+++ b/repo/packages/S/spark/2/config.json
@@ -1,0 +1,74 @@
+{
+  "type": "object",
+  "properties": {
+    "mesos": {
+      "description": "Mesos specific configuration properties",
+      "type": "object",
+      "properties": {
+        "master": {
+          "default": "mesos://zk://master.mesos:2181/mesos",
+          "description": "The URL of the Mesos master. The format is a comma-delimited list of hosts like zk://host1:port,host2:port/mesos. If using ZooKeeper, pay particular attention to the leading zk:// and trailing /mesos! If not using ZooKeeper, standard host:port patterns, like localhost:5050 or 10.0.0.5:5050,10.0.0.6:5050, are also acceptable.",
+          "type": "string"
+        }
+      },
+      "required": [
+        "master"
+      ]
+    },
+    "spark": {
+      "description": "Spark Framework Configuration Properties",
+      "type": "object",
+      "properties": {
+        "framework-name": {
+            "description": "The name of the framework. Until this is configurable, please do not change this from it's default value.",
+            "type": "string",
+            "default": "spark"
+        },
+        "host": {
+          "type": "string",
+          "description": "The host that the Spark cluster dispatcher will register under.",
+          "default": "$HOST"
+        },
+        "cpus": {
+          "default": 1,
+          "description": "CPU shares to allocate to each Marathon instance.",
+          "minimum": 0.0,
+          "type": "number"
+        },
+        "instances": {
+          "default": 1,
+          "description": "Number of Marathon instances to run.",
+          "minimum": 0,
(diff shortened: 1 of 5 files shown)


repo/packages/S/spark/2/config.json
```diff
@@ -0,0 +1,86 @@
...
+          "default": 1024.0,
+          "description": "Memory (MB) to allocate to each Marathon task.",
+          "minimum": 1024.0,
+          "type": "number"
+        },
+        "mesos-authentication-principal": {
+          "description": "The Mesos principal used for authentication.",
+          "type": "string"
+        },
+        "mesos-authentication-secret-file": {
```
tnachen: Where is this being used?

repo/packages/S/spark/2/config.json
```diff
@@ -0,0 +1,86 @@
...
+          "description": "The path to the Mesos secret file containing the authentication secret.",
+          "type": "string"
+        },
+        "mesos-role": {
+          "description": "Mesos role for this framework.",
+          "type": "string"
+        },
+        "mesos-user": {
```
tnachen: Where is this being used?

tnachen:
Btw you updated the Spark cli to look up image based on universe or?

repo/packages/S/spark/2/config.json
mgummelt: @kensipe is the author.

mgummelt:
No

repo/packages/S/spark/2/config.json
elingg: Oh, I think @tnachen's point is that this is not being used by marathon.json. It should be passed as an optional argument if we have it here.

elingg:
Can you run the build.sh script under the scripts directory so that the build will pass?

elingg:
Thanks for submitting this PR, @mgummelt!

repo/packages/S/spark/2/config.json
elingg: Here is an example of passing an optional argument : {{#chronos.cassandra-keyspace}} --cassandra_keyspace {{chronos.cassandra-keyspace}}  See https://github.com/mesosphere/universe/blob/version-1.x/repo/packages/C/chronos/1/marathon.json#L5

mgummelt: I know.  I'm saying I don't know where this is being used, nor why it's included as an option.  We'd have to ask the author, @kensipe.

What do you mean "passed as an optional argument"?  It's already not listed as "required".

mgummelt:
@elingg I'll run build.sh Can you add the build.sh instructions to the README or to a CONTRIBUTERS file?

elingg:
@mgummelt, yes, we should update the readme (oddly enough I thought these instructions were there). I'm a bit busy right now, but if you could help with that, it would be great. Or as a minimum, if you could create a Github issue explaining that this is not clear so we make note of the issue. This is definitely a pain point for users.

repo/packages/S/spark/2/config.json
elingg: Yes, it might be helpful to double check with @kensipe. Or if you know how these command line params are used, you can resolve the issue yourself by updating the marathon.json. We would like to avoid including unused options.

elingg:
One comment about the unused options above that should be addressed, but other than that looks great! Feel free to merge once that comment is addressed if there are no other objections from @tnachen

tnachen:
@elingg everything else LGTM! Thanks for asking.

mgummelt:
I've spoken with @kensipe, and we've agreed to remove the three unused config vars, which all were related to authentication.

PR #244, opened 2015-10-07
updated tags.  removed common tags and added tags that made sense

kensipe:
tags are used for search... they need to help the user discover services.

### DIFF
--- a/repo/packages/M/marathon/5/package.json
+++ b/repo/packages/M/marathon/5/package.json
@@ -16,7 +16,7 @@
       "url": "https://github.com/mesosphere/marathon/blob/master/LICENSE"
     }
   ],
-  "tags": ["mesosphere", "framework"],
+  "tags": ["init","long-running"],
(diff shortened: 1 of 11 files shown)


kensipe:
@elingg looking for a review.  this is an unusual update to existing packages

repo/packages/M/marathon/5/package.json
```diff
@@ -16,7 +16,7 @@
       "url": "https://github.com/mesosphere/marathon/blob/master/LICENSE"
     }
   ],
-  "tags": ["mesosphere", "framework"],
+  "tags": ["init","long-running"],
```
elingg: nit need a space here

elingg:
Took a quick look and it looks good to me. We would need to make sure the tags are matching.

kensipe:
what do you mean on making sure the tags are matching?

repo/packages/M/marathon/5/package.json
kensipe: good eye!

PR #245, opened 2015-10-08
spark 1.5.0 with patch for multiple roles

mgummelt:
This is a build of this commit: https://github.com/mesosphere/spark/commit/db83ac7246ed6fd183974eaf412847ac0c26b1ea , which is @tnachen's multi role fix cherry-picked onto 1.5.0

I've only tested it against our basic test suite, not against anything testing the specific fix.  @tnachen says he means to write a unit test, but we don't have that yet.

### DIFF
--- a/repo/meta/index.json
+++ b/repo/meta/index.json
@@ -111,7 +111,7 @@
       }
     },
     {
-      "currentVersion":"1.5.0",
+      "currentVersion":"1.5.0-db83ac7",
       "description":"Spark is a fast and general cluster computing system for Big Data",
       "framework":true,
       "name":"spark",
@@ -124,7 +124,8 @@
       "versions":{
         "1.4.0-SNAPSHOT":"0",
         "1.4.1":"1",
-        "1.5.0":"2"
+        "1.5.0":"2",
+        "1.5.0-db83ac7":"3"
       }
     }
   ],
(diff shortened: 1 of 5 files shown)


elingg:
The build needs to pass. We need to update repo/meta/index.json with the corresponding version. Sanity tests are okay for now, but if you can try writing a unit test or integration test  of @tnachen that will be a helpful learning experience and will give us more confidence. Testing multiple roles on a cluster would be another option.

mgummelt:
Ran scripts/build.sh.  Tests are passing now.

I'll plan on leaving this here, and writing a test for roles after I set up the spark automated build.  If you want to give cisco this build before that, let me know and I'll push.

elingg:
@mgummelt Another thing I thought of is that we want to have a level of confidence before going back to our customers so it will probably make the most sense to test this out before merge.

kensipe:
@mgummelt do you need another tester on this?

mgummelt:
tim said he would write the unit tests today, so hopefully not

elingg:
The build is still failing on this, see https://teamcity.mesosphere.io/viewLog.html?buildId=60245&buildTypeId=Oss_Universe_Ci&tab=buildLog#_focus=202. It shows a problem with the index. Could you try running scripts/build.sh? @mgummelt, the highest priority right now should be testing this to send it to Cisco before we create an automated build. @kensipe or @gabrielhartmann, that would be great if you can jump in and help with the testing here. I will see if I can take a look as well if there is time on Monday. Thanks!

mgummelt:
@elingg @gabrielhartmann plz review

I've tested this against a raw mesos cluster to validate the fix described here: https://github.com/apache/spark/pull/8872

I've also run dcos-tests against a DCOS cluster.

elingg:
LGTM @mgummelt. If no objections soon from @gabrielhartmann, let's merge!

PR #246, opened 2015-10-09
Add checks for duplicated keys to validation

discordianfish:
I wanted to do the same for marathon.json but that turned out to be more complicated given that the template first needs to be rendered.


kensipe:
nice... checking out now

kensipe:
I like it!!  thanks!

PR #251, opened 2015-10-16
Spark: the dcos proxy is based on the framework-name

kensipe:


kensipe:
@mgummelt  or @gabrielhartmann could you review?   I've tested this working in dcos.

gabrielhartmann:
Looks fine to me.  So the config value was there, just not being used anywhere?  I looked around, but didn't see it being used anywhere.

mgummelt:
"Until this is configurable, please do not change this from it's default value."

We should remove that from config.json

PR #254, opened 2015-11-04
moving memsql to the multiverse

kensipe:


kensipe:
@elingg @gabrielhartmann looking for a review

kensipe:
it appears I referenced this as mysql instead of memsql.  please forgive.  I can redo it if you like.  I think it is just the branch which is an artifact that will not be maintained.

elingg:
I'm not too picky about the branch name. LGTM

PR #255, opened 2015-11-05
Improving PreInstallNotes for HDFS

jgarcia-mesosphere:
Changing the preinstall note language from "recommend" to "require" to emphasize to users that this is a hard requirement.


kensipe:
thanks @jgarcia-mesosphere!

PR #256, opened 2015-11-06
Update URLs to use https://downloads.mesosphere.com. Fixes #171

### LINKED ISSUE #171: Downloads from downloads.mesosphere.* using wrong URL
All things which download stuff from downloadse.mesosphere.com should use the url: `https://downloads.mesosphere.com`.

Old URLs that are bad:
s3-direct: `https://s3.amazonaws.com/*`
Old domain: `https://downloads.mesosphere.io`
Non-https: `http://*.mesosphere.{io,com}`

ssk2:
I've tested a handful of these to make sure the assets are still downloadable. 

For the most part I had to move no files - but there are a bunch of Spark files in the downloads.mesosphere.com bucket under "assets". I had to move the "log4j.properties" file over into the downloads.mesosphere.io bucket under "spark/assets" since it turns out https://downloads.mesosphere.com actually points to the downloads.mesosphere.io bucket. The other files still need to be moved and the [Spark tutorial](https://docs.mesosphere.com/services/spark/) updated.

@cmaloney, @jsancio - please sanity check! An alternative approach here is to submit new versions of the packages with the URLs updated and then hope package maintainers use the new URLs going forwards. This might be preferable to updating old packages too. Let me know.


jsancio:
@ssk2 Why are we doing this? Does this break already deployed/installed packages? Meaning this doesn't change the URL in Marathon AppDefinition that have already been deployed.

Also, package revision are immutable and should not been changed after they have been deployed. Please explain why we are doing this. Is there a Jira issue for this?

ssk2:
Please see the linked issue #171 for a description. We have generally
inconsistent ways of accessing assets for packages and this can cause
breakages / confusion.

I'm unsure how it could possibly break already deployed packages - am I
missing something?

I'll create new packages with the new URLs, as per your other comment.

ssk2:
@jsancio - updated! I created new revisions for each of the packages that contained old/bad URLs and updated index.json.

ssk2:
Links fixed in documentation in https://github.com/mesosphere/mesosphere-docs/pull/223.

mgummelt:
The spark package files are now generated from: https://github.com/mesosphere/spark-build/tree/master/package

This is so we can create a package artifact from our automated spark test.  If you are going to make changes to spark, please do so there, as these won't persist.

ssk2:
@mgummelt: I've moved the Spark package changes over to `spark-build` and raised a PR there. If no-one has any objections to the other changes in this PR, I will merge this tomorrow.

PR #267, opened 2015-11-30
[kafka] Specify heap size, fix memory overhead.

brndnmtthws:
With the current default memory settings, the Kafka scheduler will get OOM killed when you try to do too much with it. This PR adds a) an option to specify the heap size and b) increases the memory requirement such that our scheduler doesn't get OOM killed.

cc @joestein


kensipe:
reviewing

kensipe:
LGTM.. limited testing but looks good.

joestein:
thnx!

Issue #252, opened 2015-10-19
Marathon package should construct zk url using marathon.zk and marathon.framework-name


jsancio:
To guarantee that the Marathon zk node is unique we should construct the url as follow: `--zk {{marathon.zk}}/{{marathon.framework-name}}`. We were doing this in revision 2 of the package but was changed when we moved to revision 3. Does anyone remember why?

cc @kolloch @aquamatthias

ssk2:
@jsancio - it looks like this might be a bug. @aquamatthias is working on a tool to generate Marathon DCOS app configs and we can re-assess this with that.

jgarcia-mesosphere:
Sorry team, just couldn't bear to wait: https://github.com/mesosphere/universe/pull/273

aquamatthias:
This is fixed in #268

jgarcia-mesosphere:
#268 is not yet available, and users are reporting issues while trying to follow the official docs. We need to merge this to make it work until #268 is available.

jsancio:
I just did an install of marathon in `version-2.x` and by default it took over the `universe` znode. It should do that ideally by default is should use `universe/marathon-user`

cc @aquamatthias

aquamatthias:
@jsancio what is the story about version 2? When will it be the default? We currently only maintain version 1.

aquamatthias:
As far as I see you created version 2

jsancio:
@aquamatthias version-2.x will become the default after 1.6.1 (when we release Cosmos). I created version-2.x for Marathon for you. I did this by copying the latest Marathon package and refactoring it to conform to the version-2.x spec.

jsancio:
Assigning back to you. Let me know if you need anything from me.

aquamatthias:
So what is the schedule for 1.6.1?

PR #268, opened 2015-12-02
Use Marathon 0.13.0

aquamatthias:
See release info here: https://github.com/mesosphere/marathon/releases/tag/v0.13.0


aquamatthias:
Fixes #252 

The user can configure the zk path. If he does, that value will be taken - we will not mess with configuration, as the ticket suggests. Only if the user does not supply a config value, a default is constructed by using zk://master.mesos:2181/universe/{framework-name}

See snippet from command line:
```
--zk {{#command-line.zk}}{{command-line.zk}}{{/command-line.zk}}{{^command-line.zk}}zk://master.mesos:2181/universe/{{command-line.framework-name}}{{/command-line.zk}}
```

repo/packages/M/marathon/8/config.json
```diff
@@ -0,0 +1,378 @@
+{
+  "additionalProperties" : false,
+  "description" : "Marathon Universe properties",
+  "properties" : {
+    "marathon" : {
+      "additionalProperties" : false,
+      "description" : "Marathon app configuration properties.",
+      "properties" : {
+        "id" : {
+          "description" : "The ID of this marathon application name.",
```
ssk2: `marathon` should be `Marathon`. This description is a little confusing. How is this different to `framework-name`? We should stick to just using `framework-name` if possible since that would be consistent with our other packages.

jsancio:
Thank you @aquamatthias! Can we make sure that we also submit a PR against the version-2.x branch. We are going to lock this branch once we release the latest DCOS CLI. Thanks!

ssk2:
@jsancio - will you test installation of this package or should I do it?

aquamatthias:
@jsancio sure. can you point me to the changes/documentation for v2?

jsancio:
@ssk2 Please test it. I don't think I will be able to get to it today. Thanks!

sttts:
It starts up and is green eventually. But in the DCOS ui there is neither an icon, nor a health state:

bildschirmfoto 2015-12-21 um 11 44 11

sttts:
Smoke tested to run a non-Docker `sleep 120` and a dockerized `echo Hello World` successfully via the **UI**.

sttts:
After setting `marathon.url` manually in the dcos-cli config (compare https://github.com/mesosphere/dcos-cli/issues/396), the cli seems to work:
```
dcos marathon task list
APP           HEALTHY          STARTED              HOST     ID
/test-docker    True   2015-12-21T11:09:40.608Z  10.0.2.175  test-docker.5442cc55-a7d3-11e5-9168-4670fbe3ae43
```

sttts:
Smoke test are successful. Not sure about the icon/health issue. Somebody from the team has to judge whether this is normal.

aquamatthias:
Thanks @sttts there is a health check in the marathon.json as well as icons in the package.json. 
@rcorral: are there any changes in the DCOS-UI I am not aware of? Why is the health state not shown? Why is the icon not visible?

ssk2:
Matthias,  you must have a property in the config called framework-name for
the health checks and icons to work in DCOS UI. I ran into the same problem
with Jenkins (see our multiverse package to see the fix).

ssk2:
@aquamatthias - I just spoke to @mlunoe. `framework-name` should be nested under an object with the same name as the package for it to be picked up correctly. In this case, it worked when it used to live under `marathon` but will not work under `command-line`. This should fix the healthchecks issue.

This leads me to a secondary point. The new config format is more confusing than the old one. Specifically, the `command-line` object. If I understand it correctly, `command-line` exposes the properties that are passed to the Marathon application via the command line to start it. However, this exposes an implementation detail of the Marathon Universe package to the end user. Personally, I'd advocate having a more logical grouping, perhaps something like:
* `framework` (properties related to the Mesos API)
* `networking` (networking, HTTP server configuration)
* `security` (SSL properties, etc.)
* `admin` (logging, Datadog properties)
(cc @jgarcia-mesosphere)

aquamatthias:
Since version 8 is already used, I created another PR #277.

aquamatthias:
@ssk2: I did not know about the requirements for DCOS UI. Where is that written down?

PR #273, opened 2015-12-18
Fix ZKnode collisions in Marathon v8

jgarcia-mesosphere:
In the Marathon package, we define `zk://master.mesos:2181/universe` without using any specifier for the ZK node. This results in ZK node name collisions for users trying to follow the [instructions here](https://docs.mesosphere.com/services/marathon/).

Let's use the `framework-name` as the ZK node. This will have the added advantage of making it easier to find in Exhibitor.

Ref: https://github.com/mesosphere/universe/issues/252


aquamatthias:
This is fixed in #268

jgarcia-mesosphere:
#268 isn't available yet, meanwhile users keep stumbling into this. Please merge this while we're waiting for #268.

PR #274, opened 2015-12-21
Document and implement specification changes

jsancio:


tamarrow:
@jsancio looks good. should we also update the package validation with the new schema?

jsancio:
@tamarrow Done. I changed the package-schema.json to only allow packages with version `"2.0"`.

repo/meta/schema/package-schema.json
```diff
@@ -68,6 +72,7 @@
     }
   },
```
tamarrow: should we also remove `images` from this schema since they should now be in `resource.json`?

tamarrow:
@jsancio thanks. can we also add a schema for `resource.json` in `repo/meta/schema/resource-schema.json`?

repo/meta/schema/package-schema.json
jsancio: Yes. We already removed it. It is part of the version-2 branch. This PR is against version-2. I am fixing the documentation and adding a `packagingVersion` property.

jsancio:
> @jsancio thanks. can we also add a schema for resource.json in repo/meta/schema/resource-schema.json?

We did. It is already part of version-2. This is a PR against that branch.

PR #281, opened 2016-01-08
Do not use hidden flags, that can not be defined in the config.

aquamatthias:
Hidden flags are not put into the config (because they are hidden).
We decided to render hidden flags into the marathon start command, in order to possibly override those.
This leads to the fact that, that hidden boolean parameters get disabled, since the config is not existent.
Example: revive-offers-for-new-apps was disabled by default, which leads to long deployments.

This fix removes all hidden parameter from the command line.


repo/packages/M/marathon/9/marathon.json
```diff
@@ -29,13 +29,13 @@
   "container": {
     "type": "DOCKER",
     "docker": {
-      "image": "mesosphere/marathon:v0.13.0",
+      "image": "mesosphere/marathon:v0.14.0-RC2",
```
gkleiman: This PR is for universe, so I guess we want to keep v0.13.0.

aquamatthias: Yes. Thanks!

PR #288, opened 2016-01-19
Marathon v0.14.0

gkleiman:
cc/ @aquamatthias


ssk2:
Thanks @gkleiman. I think we might need to add this to version-2.x as well, but perhaps @jsancio can help here. I think he has some tooling to help with translating the package and we should probably build that into the Marathon tool you use to generate these packages.

PR #304, opened 2016-02-19
removing cli for hdfs and cassandra

kensipe:


jdef:
where is the CLI moving to? do we have a new pattern that we're adopting
for integrating CLI into packages?

jsancio:
Please create a pr against version-2.x also.

kensipe:
@jsancio sure thing... should it be in the v2 format?  or can I use this existing format which can be updated later?

kensipe:
@jdef I'm not sure... I haven't been including in the loop.   The CLI are broken or useless at this point and we want to remove them.   We will have a `good` CLI replacement in the near future.

kensipe:
@jsancio forget my last comment regarding v2.  I can use the existing work you did with hdfs to create a v2 version without the cli... I will have it shortly.

BenWhitehead:
please retarget this PR at branch `version-2.x`

kensipe:
removed from `version-2.x` already in place... closing this PR

PR #305, opened 2016-02-22
Cli removal v2

kensipe:
remove of hdfs and cassandra cli from v2
I also upgraded hdfs to the current released version on v1


kensipe:
@jsancio and @keithchambers updates on version-2.x are here.   I will run a few tests but this looks good.

PR #306, opened 2016-02-22
Cli removal v2

kensipe:
removing cli from cassandra and hdfs

### DIFF
--- a/repo/meta/index.json
+++ b/repo/meta/index.json
@@ -44,7 +44,7 @@
       }
     },
     {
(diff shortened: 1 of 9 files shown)


kensipe:
@jsancio first... I screwed up on my last PR.  I thought I was up against my forked repo when I was doing my tests... and I was not.   So I reverted the branch back to its original position.  second.. cassandra looks good but there is an issue with the hdfs.  let me do some more testing.

kensipe:
@jsancio ok.. this is ready to go

PR #310, opened 2016-02-23
Increase failover timeout

m0ppers:
Current setting will easily kill the whole arangodb cluster (slave going down etc.).


joerg84:
LGTM

PR #311, opened 2016-02-24
Ported changed failover timeout from v1

joerg84:


joerg84:
This is just a bugfix with minor config change and hence does not require a new version.

PR #315, opened 2016-02-29
New Jenkins package

ssk2:
Bumped this to the latest version of the Docker image and added an optional constraint parameter. Tested this and it works. One of these apps cannot satisfy the constraint, the other is deploying to a specific node:
image

### DIFF
--- a/repo/packages/J/jenkins/3/config.json
+++ b/repo/packages/J/jenkins/3/config.json
@@ -1,0 +1,70 @@
+{
+  "type": "object",
+  "properties": {
+    "jenkins": {
+      "description": "Jenkins framework configuration properties",
+      "type": "object",
+      "properties": {
+        "framework-name": {
+          "description": "The framework name to register with Mesos.",
+          "type": "string",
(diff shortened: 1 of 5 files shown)


repo/packages/J/jenkins/3/config.json
```diff
@@ -0,0 +1,70 @@
...
+            "description": "Optional arguments to pass to the JVM.",
+            "type": "string",
+            "default": "-Xms1024m -Xmx1024m"
+        },
+        "known-hosts": {
+            "description": "A space-separated list of hosts used to populate the SSH known hosts file on the Jenkins master.",
+            "type": "string",
+            "default": "github.com"
+        },
+        "hostname": {
```
rji: Could we make this a more generic "constraints" parameter instead? Adding `hostname` when we already have `host` seems confusing to me, especially when they do two very different things.

ssk2:
@rji - addressed!

rji:
Thanks @ssk2! Would you mind squashing these commits?

PR #321, opened 2016-03-08
Fix README to reflect the latest architecture.

jsancio:


jsancio:
@tamarrow Please review. Thanks!

tamarrow:
@jsancio one question, rest LGTM

jsancio:
@tamarrow, I removed the Marathon schema file because we don't currently use it. We delegate the validation to Marathon. We don't do the validation because Marathon is always changes what is a valid Marathon AppDefinition.

PR #336, opened 2016-03-18
Update images and fix required options

m0ppers:


joerg84:
This fixes a problem for the installer with required empty default strings. 
Tested on Stable.
LGTM

PR #345, opened 2016-03-29
fix hdfs

mgummelt:
trailing commas are illegal in JSON.  This fails to validate in Cosmos, and therefore can't install.

cc @kensipe


mgummelt:
subsumed by https://github.com/mesosphere/universe/pull/346

PR #354, opened 2016-04-01
Packages should manage their own framework name

jsancio:
The latest version of Cosmos is going to force packages to manage their
own framwwork name. This change makes it possible for us to change
Cosmos without breaking all of the packages.

PR #365, opened 2016-04-06
reduce default memory required for jenkins

rji:
This commit reduces the default (minimum) amount of memory for the
Jenkins service from 2 GB to 1 GB, and modifies the JVM heap size from
1024m to 512m.


jsancio:
@rji Let's create a new package revision. Technically, package are immutable.

ssk2:
@jsancio - but this commit didn't create new packages - https://github.com/mesosphere/universe/pull/354/files

jsancio:
Rules are meant to be broken.

rji:
Updated PR to make this change in a new package revision.

PR #384, opened 2016-04-10
Update Chronos package so framework-name and ID are the same property.

ssk2:
Previously, the Chronos package would use a separate ID and framework-name config parameter to set the app ID and package label appropriately. This PR creates a new package without the ID parameter, to be consistent with all of our other packages.

CC @mhausenblas.


BenWhitehead:
please open a new pr targeted at `version-3.x` if this change should still be included

PR #403, opened 2016-04-16
support for Avi-Vantage package, address review comments, move to beta image

ksivask:


kensipe:
@ksivask attempts to install fail

PR #405, opened 2016-04-18
Fix Storm package notes

guenter:


kensipe:
@guenter your changes are text only and could not break the build.  I think this is an auth issue and will take a look

kensipe:
lgtm

there are other issues (using mesosphere in the tags doesn't make sense for example... but they are separate issues I will resolve in another PR)

kensipe:
@guenter this needs to be rebased... I'll handle it. consider this merged.

PR #406, opened 2016-04-18
Nginx package

mohitsoni:
* Configurable HOST vs BRIDGE mode networking.
* Ability to host static content via NGINX
* Working health-checks
* Ability to scale both horizontally (not nginx-ha) and vertically (increase CPU and memory)
* Run multi-tenant NGINX instances on the same DC/OS cluster as well as a same host (in BRIDGE mode).


mhausenblas:
Seems there's a problem with the build @mohitsoni? Can you check the TC output pls?

mohitsoni:
@mhausenblas @pyronicide I took a look at the failure. And, it was not very clear why the build fails. Any pointers ?

PR #412, opened 2016-04-18
Add Calico as a Universe package

robbrockbank:
PR to add a Calico framework to the Universe packages.

This framework does the following:
1. Runs an etcd proxy on each agent
2. Configures Docker to use etcd as it's cluster store (thus enabling Docker multi-host networking)
3. Configures Mesos Agent to use Netmodules with a Calico plugin
4. Runs the Calico networking daemon and the Calico Docker networking driver on each Agent

This framework is a little odd in that it reconfigures the agents which will cause transient task loss on those agents.  However:
-  The framework limits the number of concurrent agent restarts.  By default it is set to 1 - so only one agent can be updated at a time.  After an agent is updated, we wait for 30s to ensure the agent process is stable before updating the next agent.
-  Once an agent is updated, it will not be updated again.

Future looking:
-  We'd like the net modules library to be installed as part of the core installation (or as an optional installation package)
-  Perhaps Docker multi host networking will be enabled on installs as part of the backlog item to support Docker user defined networks.


djosborne:
@kensipe Looking for feedback

repo/packages/C/calico/0/marathon.json.mustache
```diff
@@ -0,0 +1,55 @@
+{
+  "id": "calico",
```
kensipe: prefer that you use the "{{calico.framework-name}}" for the value.

repo/packages/C/calico/0/marathon.json.mustache
```diff
@@ -0,0 +1,55 @@
...
+    "CALICO_INSTALLER_URL": "{{resource.assets.uris.calico-installer}}",
+    "CALICO_MESOS_PLUGIN": "{{resource.assets.uris.calico-mesos-plugin}}",
+    "ETCD_SRV": "{{etcd.etcd-discovery-url}}",
+    "ETCD_BINARY_URL": "{{resource.assets.uris.etcd}}",
+    "NETMODULES_URL": "{{resource.assets.uris.netmodules-libraries}}",
+    "MESOS_MASTER": "{{mesos.master}}"
+  },
+  "labels": {
+    "DCOS_PACKAGE_FRAMEWORK_NAME": "{{calico.framework-name}}"
+  },
```
kensipe: may want to add:  "MARATHON_SINGLE_INSTANCE_APP":"true" to labels

repo/meta/index.json
```diff
@@ -18,6 +18,23 @@
       }
     },
     {
+      "currentVersion":"0.1.0",
+      "description":"Calico networking for Mesos.  Calico provides IP-per-task functionality with rich policy.  This framework optionally installs Docker multi-host networking (required for Calico with the Docker containerizer) and Mesos Agent network hooks with netmodules (required for Calico with Mesos tasks).  Both installations will cause temporary restarts of an Agent causing transient task loss.  The number of agents restarted at the same time is configurable and defaults to a single concurrent restart.  At the moment, only a limited set of Mesos versions and OSes are supported - check the documentation for details.  The framework will still run on unsupported versions, in which case the Mesos network hooks will not be installed and Calico networking will only be available with the Docker containerizer.",
```
kensipe: would rather state that this is for dcos (instead of mesos).    the underlying details we are abstracting a bit.   Also there is a preference to use the word service over framework.

repo/packages/C/calico/0/config.json
```diff
@@ -0,0 +1,169 @@
...
+      "properties": {
+        "framework-name": {
+          "type": "string",
+          "description": "The name of the framework. Until this is configurable, please do not change this from its default value.",
+          "default": "calico"
+        },
+        "calicoctl-url": {
+          "type": "string",
+          "description": "The download URL for the calicoctl command line tool.",
+          "default": "https://github.com/projectcalico/calico-containers/releases/download/v0.18.0/calicoctl"
```
kensipe: all urls should be from the resource.json file.

repo/packages/C/calico/0/marathon.json.mustache
```diff
@@ -0,0 +1,55 @@
...
+      "network": "HOST",
+      "forcePullImage": true
+    }
+  },
+  "args": [],
+  "cpus": {{calico.cpu-limit-framework}},
+  "mem": {{calico.mem-limit-framework}},
+  "instances": 1,
+  "env": {
+    "CALICO_CALICOCTL_URL": "{{calico.calicoctl-url}}",
```
kensipe: when you change the url to resource.json this variable will change.

robbrockbank: @kensipe   This actual line of code remains unchanged though - correct?

We do want the environment set from the calico.calicoctl-url because that field is configurable ... just that it defaults to the value in the resource assets.

repo/meta/index.json
robbrockbank: ok

repo/packages/C/calico/0/config.json
robbrockbank: ah yup - missed this one

repo/packages/C/calico/0/marathon.json.mustache
robbrockbank: yup will do

repo/meta/index.json
```diff
@@ -18,6 +18,23 @@
       }
     },
     {
+      "currentVersion":"0.1.0",
+      "description":"Calico networking for DCOS.  Calico provides IP-per-task functionality with rich policy.  This framework optionally installs Docker multi-host networking (required for Calico with the Docker containerizer) and Mesos Agent network hooks with netmodules (required for Calico with Mesos tasks).  Both installations will cause temporary restarts of an Agent causing transient task loss.  The number of agents restarted at the same time is configurable and defaults to a single concurrent restart.  At the moment, only a limited set of Mesos versions and OSes are supported - check the documentation for details.  The framework will still run on unsupported versions, in which case the Mesos network hooks will not be installed and Calico networking will only be available with the Docker containerizer.",
```
robbrockbank: Should I remove "Mesos" all together?

repo/packages/C/calico/0/marathon.json.mustache
djosborne: @robbrockbank No, this needs a slight change to: `"{{resource.assets.uris.calicoctl-url}}"`

repo/meta/index.json
kensipe: that would be great

repo/packages/C/calico/0/config.json
```diff
@@ -0,0 +1,169 @@
...
+      "properties": {
+        "framework-name": {
+          "type": "string",
+          "description": "The name of the framework.",
+          "default": "calico"
+        },
+        "calicoctl-url": {
+          "type": "string",
+          "description": "The download URL for the calicoctl command line tool.",
+          "default": "{{resource.assets.uris.calicoctl}}"
```
kensipe: this is incorrect and is causing a failure on installation... the calioctrl-url property in this file should not exist at all.

repo/packages/C/calico/0/marathon.json.mustache
robbrockbank: Ah I gotcha now - I assume then we also need to move the calico0node and calico-libnetwork images to the resources as well - since these are container images

repo/packages/C/calico/0/config.json
robbrockbank: Yup - sorry didn't spot that!  Removed this and the two container images below - I've moved these to be assets too.

kensipe:
tested with some failing tests but reaches a point of stability.

kensipe:
awesome!!  thanks!

PR #413, opened 2016-04-18
update post install notes

mgummelt:
@keithchambers PTAL


kensipe:
@mgummelt there is a conflict... I think the tags changed so it can't merge.

PR #423, opened 2016-04-19
remove virtual ip labels

jfrazelle:


kensipe:
@jfrazelle changes like this require a new index... (which are hard to diff, but are expected by cosmos).  so this change in `mysql/0` should result in a `mysql/`` with change.   don't forget to rerun the index scripts.

jfrazelle:
hmm i did but maybe on an old checkout

jfrazelle:
updated :)

jfrazelle:
wow no sleep is killing me, now its good

kensipe:
@jfrazelle perfect!

kensipe:
this service fails to install

```

jfrazelle:
ok we need to bump the memory for some reason on mine it doesnt need that much, will update thanks so much

jfrazelle:
ok i have bumped to 512

kensipe:
@jfrazelle I had the same failure... I bumped mem to 1024 and cpu to 1 and got it running.  I would add that the service doesn't recover well.  multiple attempts with failures resulted `/tmp/mysql` being established... then when a "good" config worked it failed because the dir existed.

Let me know what you want to do about the resource needs.  It isn't clear to me what the real needs of the service are.

jfrazelle:
Hmmm ya it's not exactly clear to me either what the resource needs I upped the values though :/

kensipe:
LGTM  merging!

jfrazelle:
thanks!

PR #425, opened 2016-04-20
New Confluent release to fix config copy/paste bug

gabrielhartmann:
New release to fix config copy/paste bug


gabrielhartmann:
Tested on a cluster.  All is as expected.  Corresponded with David from Confluent and he agreed to the new versioning scheme.

PR #435, opened 2016-04-27
Fixing docker image reference.

egorpushkin:
Fixes this error caused by using resource path instead of actual docker image. 

```
State
TASK_FAILED
Message
Failed to launch container: Failed to 'docker -H unix:///var/run/docker.sock pull resource.assets.container.docker.hue-docker:latest': exit status = exited with status 1 stderr = Error: image library/resource.assets.container.docker.hue-docker:latest not found
```

### DIFF
--- a/repo/packages/H/hue/0/marathon.json.mustache
+++ b/repo/packages/H/hue/0/marathon.json.mustache
@@ -3,7 +3,7 @@
     "container": {
         "type": "DOCKER",
         "docker": {
-            "image": "resource.assets.container.docker.hue-docker",
+            "image": "{{resource.assets.container.docker.hue-docker}}",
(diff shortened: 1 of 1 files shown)


mhausenblas:
Can I please ask you to re-submit this PR in ca. one week time once we've the new version of the Universe in place? If you've got any questions, happy to discuss it via http://chat.dcos.io

egorpushkin:
@mhausenblas The fix it pretty trivial. Please consider applying it in the way that works best for you.

mhausenblas:
OK, thanks @egorpushkin — background is that we're not on `version-3.x` and GitHub doesn't allow PRs being carried over, so …

PR #441, opened 2016-05-03
[exhibitor] Add persistent volume.

brndnmtthws:
Proof:

screen shot 2016-05-03 at 6 26 06 pm

screen shot 2016-05-03 at 6 28 41 pm

### DIFF
--- a/repo/packages/E/exhibitor/1/config.json
+++ b/repo/packages/E/exhibitor/1/config.json
@@ -1,0 +1,101 @@
+{
+  "properties": {
+    "exhibitor": {
+      "additionalProperties": false,
+      "description": "Exhibitor specific configuration properties",
+      "properties": {
+        "app-id": {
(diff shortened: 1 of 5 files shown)


repo/packages/E/exhibitor/1/config.json
```diff
@@ -0,0 +1,101 @@
...
+          "type": "string"
+        },
+        "mem": {
+          "default": 3072.0,
+          "description": "Memory (MiB) to allocate to each Exhibitor instance.",
+          "minimum": 1024.0,
+          "type": "number"
+        },
+        "volume-size": {
+          "default": 10240,
```
jdef: this seems large. do we really expect 10GB of data here in large deployments?

brndnmtthws: 10G isn't very much. Better safe than sorry.

PR #442, opened 2016-05-06
Add single instance app property

ssk2:
Added `MARATHON_SINGLE_INSTANCE_APP` - which prevents users from scaling this beyond one instance. It was also necessary to add an upgradeStrategy.

Tested and working - right now the error message isn't too intuitive but will be improved with future Marathon versions:
image


jdef:
@sunil can you file a marathon ticket for that informative error message?

ssk2:
@jdef - I raised it with the Marathon UI team and they have a related ticket already :)

PR #448, opened 2016-05-10
update linkerd and namerd packages to 0.5.0

siggy:
https://github.com/BuoyantIO/linkerd/releases/tag/0.5.0

### DIFF
--- a/repo/packages/L/linkerd/2/marathon.json.mustache
+++ b/repo/packages/L/linkerd/2/marathon.json.mustache
@@ -1,0 +1,54 @@
+{
+  "id": "{{linkerd.name}}",
+  "instances": {{linkerd.instances}},
+  "cpus": {{linkerd.cpus}},
+  "mem": {{linkerd.mem}},
+  "acceptedResourceRoles": ["*", "slave_public"],
+  "constraints": [["hostname", "UNIQUE"]],
+  "maintainer": "hello@buoyant.io",
+  "container": {
+    "type": "DOCKER",
+    "docker": {
+      "image": "{{resource.assets.container.docker.linkerd-docker}}",
+      "network": "BRIDGE",
+      "privileged": true,
+      "portMappings": [
+        {
+          "containerPort": {{linkerd.admin-port}},
+          "hostPort": {{linkerd.admin-port}},
+          "servicePort": 0,
+          "protocol": "tcp"
+        },
+        {
+          "containerPort": {{linkerd.routing-port}},
+          "hostPort": {{linkerd.routing-port}},
+          "servicePort": 0,
+          "protocol": "tcp"
+        }
+      ]
+    }
+  },
+  "healthChecks": [
+    {
+      "protocol": "HTTP",
+      "portIndex": 0,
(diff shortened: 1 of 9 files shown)


repo/packages/L/linkerd/2/marathon.json.mustache
```diff
@@ -0,0 +1,54 @@
...
+  "args": [
+    "/mnt/mesos/sandbox/{{linkerd.config-filename}}"
+  ],
+  "requirePorts":true,
+  "ports": [
+    {{linkerd.admin-port}},
+    {{linkerd.routing-port}}
+  ],
+  "uris": [
+    "{{linkerd.config-uri}}/{{linkerd.config-filename}}"
```
jsancio: We should express this in the `resource.json` file. See https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/C/cassandra/1/resource.json#L4-L6 and https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/C/cassandra/1/marathon.json.mustache#L10-L13

repo/packages/N/namerd/2/marathon.json.mustache
```diff
@@ -0,0 +1,61 @@
...
+    "/mnt/mesos/sandbox/{{namerd.config-filename}}"
+  ],
+  "requirePorts":true,
+  "ports": [
+    {{namerd.admin-port}},
+    {{namerd.http-port}},
+    {{namerd.thrift-port}}
+  ],
+  "uris": [
+    "{{namerd.config-uri}}/{{namerd.config-filename}}"
```
jsancio: We should express this in the `resource.json` file. See https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/C/cassandra/1/resource.json#L4-L6 and https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/C/cassandra/1/marathon.json.mustache#L10-L13

repo/packages/L/linkerd/2/marathon.json.mustache
siggy: Hmm, we need these to be settable by the user. I didn't think that was possible via `resource.json`.

repo/packages/N/namerd/2/marathon.json.mustache
siggy: (same comment as above)

repo/packages/L/linkerd/2/marathon.json.mustache
jsancio: In general user configurable value should be passed to the task by either command line flags or environment variables. What some packages do is grab those environment variable and command line flags and create the configuration files before launching the long lived process.

The general issue with configuration through files/uri is that they need to get hosted so that the Mesos Agent can download them through HTTP. Universe and Cosmos currently don't support this functionality so it complicate deployment for our users.

siggy: All makes sense. Per offline convo, we'll continue with this implementation until we can leverage some kind of feature in DC/OS for managing user-supplied config files.

PR #449, opened 2016-05-12
Added Jenkins version to service package version

tommysdk:
Added the Jenkins version to the service package version since there are many different Jenkins versions available with different characteristics (JDK support, plugin compatibility etc) and it would be useful to know which Jenkins version to install when selecting a Jenkins package in DC/OS. With current master, there is no obvious way to know the Jenkins version beforehand.

This also aligns with other packages that uses the same denotation.


mhausenblas:
@ssk2 can you pls review this PR? Has been sitting here for almost a month :(

ssk2:
Oops, apologies @tommysdk. I can't merge this into the current package because we're about to significantly refactor this package (next week) but I will make sure this change happens.

tommysdk:
@ssk2 ok. Would appreciate a notification when the change has been introduced!

ssk2:
@tommysdk, we are now doing this: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/J/jenkins/11/package.json#L4

tommysdk:
@ssk2 nice, thanks!

PR #452, opened 2016-05-13
Updating local-universe.py to use a docker container.

pyronicide:
This changes the local-universe distribution method from a tarball to a self-hosting docker container. The base container (`multi`) contains both nginx and the docker registry. Registry assets are laid out on disk via. a local registry and volume. Both those assets and the http ones (images, non-docker packages) are copied into the final container `mesosphere/universe`.

The CLI interfaces have also changed slightly. `sudo` is now a requirement, `--selected` has been introduced to pull in *only* the packages that have been selected and `jsonschema` is now a requirement for running the script.


jsancio:
Very cool @pyronicide! What are your thoughts around the `mesosphere/universe-1.7` repo? How are we going to support that repository?

jdef:
this is a great idea

scripts/local-universe.py
```diff
@@ -115,38 +171,121 @@ def enumerate_docker_images(package_path):
     return (name for _, name in dockers.items())
 
 
-def download_docker_image(name, use_sudo):
+@contextlib.contextmanager
+def run_docker_registry(volume_path):
```
jsancio: For my own edification why are we running the docker registry and kill it after we are done?

scripts/local-universe.py
```diff
@@ -115,38 +171,121 @@ def enumerate_docker_images(package_path):
...
-        with tempfile.NamedTemporaryFile() as tempFile:
-            shutil.copyfileobj(response, tempFile)
-            tempFile.flush()
-            zip_file.write(tempFile.name, str(archive_path))
+def format_image_name(host, name):
+    # Probably has a hostname at the front, get rid of it.
+    if '.' in name.split(':')[0]:
+        return '{}/{}'.format(host, "/".join(name.split("/")[1:]))
 
+    return '{}/{}'.format(host, name)
```
jsancio: I think we can safely use `/` as separators for `[(host:port)]/user/name`. `/` are not valid characters for `user` and `name`.

> The image name which is made up of slash-separated name components, optionally prefixed by a registry hostname. The hostname must comply with standard DNS rules, but may not contain underscores. If a hostname is present, it may optionally be followed by a port number in the format `:8080`. If not present, the command uses Docker's public registry located at `registry-1.docker.io` by default. Name components may contain lowercase characters, digits and separators. A separator is defined as a period, one or two underscores, or one or more dashes. A name component may not start or end with a separator.

pyronicide: Not sure I understand what you're saying here? This is a `host/docker_image:tag`.

scripts/local-universe.py
pyronicide: The end result is a docker container that has a running docker registry (with all the containers pushed) and a running nginx serving up static files. I can't put `push` from a local docker to remote in a Dockerfile. So, the registry starts up with a local volume (`/tmp/foobar`). We then push all the images to the registry and shut it down. At this point in time, `/tmp/foobar` will have the layers and everything correctly laid out in it. These files then get placed into the end container via. a Dockerfile.

scripts/local-universe.py
jsancio: Right the full name is `host:port/username/imagename`. For example `docker.io/mesosphere/marathon`

scripts/local-universe.py
jsancio: Got it.

scripts/local-universe.py
pyronicide: That's not actually true. Docker hub has promoted that scheme, the registry doesn't care at all. It treats everything after the hostname as the image name (it goes in as key/value pairs). I'm taking the full container location and slapping it on the hostname. The only reason this function exists, is because some of the containers in the universe currently list a hostname and image names can't contain `.`s.

jsancio:
@pyronicide Looks like we need a rebase or merge. There are conflicts against version-2.x.

Issue #445, opened 2016-05-09
Avi package docker image in resource.json isn't fully specified


jsancio:
A package's resource.json needs to fully specify the resources used. That includes PNG, HTTP resources and docker images. E.g.:

```
> env TMPDIR='.' python3 scripts/local-universe.py --repository repo/packages/ --out-file local_universe.zip --sudo --include=avi
Adding https://s3-us-west-2.amazonaws.com/dcos/avi-small.png at avi/images/avi-small.png to file zip.
Adding https://s3-us-west-2.amazonaws.com/dcos/avi-medium.jpg at avi/images/avi-medium.jpg to file zip.
Adding https://s3-us-west-2.amazonaws.com/dcos/avi-high.jpg at avi/images/avi-high.jpg to file zip.
Pull docker images: avinetworks/controller
Using default tag: latest
Trying to pull repository docker.io/avinetworks/controller ... not found
Tag latest not found in repository docker.io/avinetworks/controller
Pull docker images: avinetworks/se
Using default tag: latest
Trying to pull repository docker.io/avinetworks/se ... not found
Tag latest not found in repository docker.io/avinetworks/se
Saving docker images: ['avinetworks/controller', 'avinetworks/se']
Error response from daemon: could not find image: no such id: docker.io/avinetworks/controller
```

If the docker image doesn't include a tag docker assumes the `latest` tag. In this case the `latest` tag doesn't exists so downloading fails. Looking at `marathon.json.mustache` it looks like the tag is specified out of band: https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/A/avi/0/marathon.json.mustache#L27.

jsancio:
cc @ksivask @kensipe

ksivask:
working on resolving this.

PR #455, opened 2016-05-15
Avi - issue #445 fix

ksivask:


repo/packages/A/avi/0/marathon.json.mustache
```diff
@@ -60,6 +61,13 @@
         "hostPath": "/var/run/docker.sock",
         "containerPath": "/var/run/docker.sock",
         "mode": "RO"
+      },
+      {
+        "containerPath": "unused",
+        "mode": "RW",
+        "persistent": {
+          "size": 10
```
jdef: is this disk truly "unused"? if so then why is it in this spec? if it's actually needed (and used to store data) then I'd recommend parameterizing the size here

ksivask:
addressed PR review comments

repo/packages/A/avi/0/config.json
```diff
@@ -101,6 +101,11 @@
               "minimum": 9161,
               "maximum": 9161
             },
+            "persistent-vol-size": {
+              "description": "Size of persistent volume",
+              "type": "integer",
+              "default": 10
```
jdef: is there a max/min that makes sense for this?

ksivask:
good point, yes min/max needed.

ksivask:
helo James, I have made the changes as per the comments. Can you please review?

pyronicide:
As you're adding functionality, please bump this package to a new version (1).

ksivask:
more of a bug fix than functionality here. the persistence was something left out before 1.7live. 
For next set of features, I will bump the versions.

PR #463, opened 2016-05-25
Concord Package for DCOS

RBlafford:


mhausenblas:
@RBlafford can you resolve the conflicts pls?

repo/meta/index.json
```diff
@@ -83,6 +83,23 @@
...
     },
     {
+      "currentVersion":"0.3.16.4",
+      "description":"High Performance Stream Processing with C++ and Apache Mesos",
+      "framework":true,
+      "name":"concord",
+      "selected":false,
+      "tags":[
+        "stream",
+        "concord",
```
jdef: I don't think `concord` is needed as a tag here since it's already the package name

repo/packages/C/concord/0/config.json
```diff
@@ -0,0 +1,51 @@
...
+	  "type": "string"
+	}
+      }
+    },
+    "concord": {
+      "description": "Concord configuration properties",
+      "type": "object",
+      "properties": {
+        "zookeeper": {
+	  "default": "zk://master.mesos:2181/concord",
```
jdef: It would be nice to make it easier for people to install multiple versions of this package on dc/os if they want to. Providing a default value here lets people shoot themselves in the foot more easier than not.

two suggestions:
(1) define a `cluster-name` setting, with a default of `concord` or some such
(2) instead of providing a default value here for `zookeeper`, you could handle the default case in the marathon JSON template, at which point you could also use the `cluster-name` setting

see cassandra for an example of this: https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/C/cassandra/1/marathon.json.mustache#L62

The idea is that when people want to install multiple versions of this package, they can just change the `cluster-name` setting without worrying about the `zookeeper` settings.

RBlafford:
@mhausenblas @jdef thanks for the input, I have implemented the suggestions and merged the latest updates from version-2.x into this branch.

mhausenblas:
@RBlafford thank you! However, as you can see from the checks status there seem to be issues on TeamCity with that build, can you double check pls that it's valid what you submitted?

RBlafford:
@mhausenblas I looked into that. It looks like there is an issue in the mainline branch as other peoples PR's are also failing for the same reason.

repo/packages/C/concord/0/marathon.json.mustache
```diff
@@ -0,0 +1,27 @@
+{
```
jdef: you probably also want a label that specifies the dcos package name, ala https://github.com/mesosphere/universe/blob/version-2.x/repo/packages/C/cassandra/1/marathon.json.mustache#L35

BenWhitehead:
the build `Universe Server Docker Image` is a new build for `version-3.x` branch, this PR is okay to merge to `version-2.x` without it passing. As long as the "Required" status checks are green it's okay to merge the PR.

RBlafford:
@mhausenblas I've addressed all of the feedback, is there anything else I can do before mesosphere is ready to merge?

mhausenblas:
Thanks a lot @RBlafford @BenWhitehead and @jdef!

mhausenblas:
So I just did another test install and came across a bug we need to fix ASAP: the service name is `dcos` which is totally confusing. Can you pls send in a new PR @RBlafford?
screen shot 2016-06-04 at 00 42 18

PR #467, opened 2016-05-27
Update schemas to enforce the difference between v2.0 packages and v3.0 packages

BenWhitehead:
If a package publisher wants to take advantage of v3.0 package features they will have to set the `packagingVersion` to `3.0` in their package.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,19 @@ The required fields are:
 - maintainer
 - description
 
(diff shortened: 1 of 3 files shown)


BenWhitehead:
@jsancio Please review

jsancio:
LGTM! :ship: it

PR #470, opened 2016-05-31
Replace DCOS with DC/OS

surdy:
Updated instances of "DCOS" to be "DC/OS" . 
This will also fix tests for https://github.com/dcos/dcos-cli/pull/637


BenWhitehead:
Updating the readme is okay, but I don't think the packages should be updated. Packages are supposed to be immutable and not change once published.

surdy:
@BenWhitehead Makes sense. I'm closing this out and will undo changes in dcos-cli as well that correspond to this.

PR #471, opened 2016-05-31
Gen universe

jsancio:


BenWhitehead:
Cosmos is returning a schema validation error when trying to read the generated `index.json` from the v2 zips.

`io.circe.DecodingFailure: String: El(DownField(0.3.0),true),El(DownField(versions),true),El(DownArray,true),El(DownField(packages),true)`

It's expecting the value of `.packages.[].versions['0.3.0']` to be `String` not `Number`

PR #472, opened 2016-06-01
Update marathon.json to use port 8085

BenWhitehead:
DC/OS Private Agents don't (by default) have 80 and 443 in their defined port resources, so marathon will never get a resources offer that can be used to run the universe server container.


jsancio:
LGTM.

PR #473, opened 2016-06-02
New container image for ArangoDB framework (bugfix).

neunhoef:
This fixes a problem with persistent volumes having to do with file ownership.


mhausenblas:
@neunhoef we're in the process of moving to a new version, can I ask you to re-submit this one after the transition is completed (mid next week). Happy to chat on Slack or clarify via mail if necessary.

PR #475, opened 2016-06-03
GeoServer HA package

gisjedi:
GeoServer HA package supports scaling both vertically and horizontally. The initial application bootstraps Geoserver instances with provided settings and ensures they remain in sync.


mhausenblas:
And same here, closing in preparation for transition to new Universe

PR #476, opened 2016-06-03
OGC Preview web application

gisjedi:
Discover, browse and export data from OGC service layers exposed by way of WMS/WFS service interfaces, with a special focus given to time-enabled data. Defaults to using an existing install of the GeoServer package in a previous pull request. Install settings can be provided to point app at an externally hosted GeoServer, as well.

### DIFF
--- a/repo/meta/index.json
+++ b/repo/meta/index.json
@@ -401,6 +401,21 @@
       }
     },
     {
+      "currentVersion":"0.10.76",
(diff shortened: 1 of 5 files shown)


mhausenblas:
@gisjedi sorry it took us that long but we're in the middle of a transition to a new version. Can I please ask you to re-submit this PR in ca. one week time once we've the new version of the Universe in place? If you've got any questions, happy to discuss it via http://chat.dcos.io

PR #477, opened 2016-06-03
PostGres single-node package.

gisjedi:
PostGres package running PostGres 9.5 with PostGIS 2.2 extensions in DCOS in a single-node layout. Future release will support DB replication to allow for multiple read-replicas.


mhausenblas:
Same here, please re-submit after transition to new version of Universe is completed.

PR #478, opened 2016-06-04
Changing default framework name to 'concord'

RBlafford:


RBlafford:
@mhausenblas nice catch, this should fix the confusing issue

PR #479, opened 2016-06-06
ArangoDB 0.4.0 supporting OAuth, binary installation

m0ppers:


BenWhitehead:
@m0ppers can you re-target this PR at the version-3.x branch? that branch is the only branch that will have schema support for the binary cli resources you are referencing in `resource.json`

Blocked until #480 merges.

PR #488, opened 2016-06-13
Ss/doc link fix

sascala:
did I miss anything that you can see, @joel-hamill ?


joel-hamill:
LGTM

BenWhitehead:
@joel-hamill @sascala What is the motivation for this change? Packages are supposed to be treated as immutable and not change once published.

joel-hamill:
@BenWhitehead the doc link is wrong. the other change (`DCOS` to `DC/OS` is optional and was bundled in this fix).

BenWhitehead:
@joel-hamill I see. We might need to start tracking the URLs used in packages, since packages will most likely not be able to be updated in the future.

joel-hamill:
@BenWhitehead i totally agree this is unsustainable long term. seems like maybe we should keep this kind of usage info in the actual universe repo.

PR #489, opened 2016-06-13
Removing the Docker Swarm package until we get more resources

jimenez:


lezzidan:
hi, is there a way to test swarm again on a small testbed?

PR #491, opened 2016-06-15
Update resource schema to not allow additional properties

BenWhitehead:


BenWhitehead:
@tamarrow please review

tamarrow:
LGTM. we should update cosmos also.

BenWhitehead:
Will update cosmos as well

PR #517, opened 2016-06-27
Support fetching Nginx config from remote source

rji:
This commit adds additional functionality to the Nginx package to fetch
a configuration file from a remote source, which is useful when needing
to configure Nginx as a reverse proxy for a Marathon application instead
of using it to host content.

This commit allows the use of two different conventions; only one may be
used at a time:
  * user provides a URL to a Nginx config `nginx`
  * user provides a URL to a Nginx config `someconfig.conf`

To use this, users need to specify a configUrl that is compatible with
the Mesos Fetcher. Note that users CANNOT specify both configUrl and
contentUrl, else they will be left with an invalid Marathon application
object.

It also bumps the Nginx version from 1.8.1 to 1.10.1.


rji:
See b91565d for a diff between v1 and v2. Happy to squash these commits post-review, or feel free to squash them in the merge.

PR #525, opened 2016-06-28
Address Adam's comments

karya0:
* Added authentication.service_account_secret_name.

### DIFF
--- a/repo/packages/S/spark/12/config.json
+++ b/repo/packages/S/spark/12/config.json
@@ -129,10 +129,10 @@
                             "description": "Mesos scheduler dcos-oauth configuration.",
                             "type": "object",
                             "properties": {
-                                "enabled": {
-                                    "description": "Whether to enable dcos-oauth.",
-                                    "type": "boolean",
-                                    "default": false
+                                "service_account_secret_name": {
+                                    "default": "",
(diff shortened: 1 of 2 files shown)


repo/packages/S/spark/12/config.json
```diff
@@ -129,10 +129,10 @@
                             "description": "Mesos scheduler dcos-oauth configuration.",
                             "type": "object",
                             "properties": {
-                                "enabled": {
-                                    "description": "Whether to enable dcos-oauth.",
-                                    "type": "boolean",
-                                    "default": false
+                                "service_account_secret_name": {
```
adam-mesos: Is this what hdfs calls it? It'd be nice to be consistent between the two if possible.

karya0: Yes, @nickbp and I agreed on this name yesterday.

adam-mesos:
:rocket:

repo/packages/S/spark/12/config.json
```diff
@@ -129,10 +129,10 @@
...
+                                    "default": "",
```
mgummelt: Let's remove this default.  An empty string is not the same as unset.

repo/packages/S/spark/12/config.json
mgummelt: Let's call this "secret_name".  It's implicit that it's for the service.

PR #529, opened 2016-06-30
Bump dcos-security-cli version

vespian:


vespian:
@tamarrow Could you please have a look and merge ? Thanks!

BenWhitehead:
please create a new revision of the package rather than changing the existing package.

vespian:
@BenWhitehead This binary has not been released yet/nobody is using it yet, and the previous versions has been destroyed anyway - I accidentally published v0.8.2 version binaries as v0.8.0 and overwritten s3 bucket.

If you still want to create new release then ok, can do :)

PR #533, opened 2016-06-30
Preserve order of properties in config.json

jsancio:
Make sure that we preserve the order of the config.json properties by
deserializing them into an OrderedDict.


tamarrow:
LGTM

ssk2:
Nice!

PR #544, opened 2016-07-06
spark 100 [WIP]

mgummelt:


jdef:
um. 100? why the large version gap? just curious...

PR #549, opened 2016-07-07
[marathon-lb] Bump to v1.3.1, fix map param.

brndnmtthws:


brndnmtthws:
Tested and validated:

screen shot 2016-07-07 at 8 17 10 am
screen shot 2016-07-07 at 8 17 40 am


😀

PR #557, opened 2016-07-13
darwin universe builds

kensipe:
now works on darwin and linux.  defaults to the assumed linux platform unless darwin is detected.

references for command equivancy:  
(linux) http://www.computerhope.com/unix/stat.htm and 
(darwin) https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/stat.1.html


kensipe:
I should add that it is tested working on darwin.

PR #563, opened 2016-07-15
Updating Elasticsearch to scheduler 1.0.1

akaplan-mesosphere:
This update removes the need for defining a discrete executor image and makes the settings location optional.  Moved to packagingVersion 3.0.


philwinder:
Awesome job @akaplan-mesosphere, thanks.

PR #580, opened 2016-07-21
[marathon-lb] Version bump to v1.3.2.

brndnmtthws:


brndnmtthws:
Tested and validated:

screen shot 2016-07-21 at 11 56 54 am
screen shot 2016-07-21 at 11 55 58 am

Hugs all 'round

PR #581, opened 2016-07-22
add script to generate config reference tables

rji:
For the Jenkins package, we maintain a configuration reference with our
documentation. Up to this point, we've generated Markdown tables with
configuration properties by hand, but the package is at the point where
it's no longer feasible for someone to do this by hand.

This commit introduces a script that walks the Universe package
repository and builds a single Markdown file with a configuration
reference. A user can run this script, get a single Markdown file in
their current working directory, and copy/paste the resulting table into
their documentation tool of choice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ target/
 .idea/*
 tmp/
 repo/packages/T/tests
+
+config-reference.md
--- a/scripts/generate-config-reference.py
+++ b/scripts/generate-config-reference.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+"""This script builds a Markdown file containing configuration references for
+all packages (and all package versions) contained in the Mesosphere DC/OS
+Universe repository. It outputs a single file, 'config-reference.md' in the
+current working directory.
+
+  Usage:  ./generate-config-reference.py [/path/to/universe/repo/packages]
+
+"""
+import json
+import os
+import sys
+
+
+def find_config_files(path):
+    config_files = []
+
+    for root, dirs, files in os.walk(path):
+        for f in files:
+            if f == 'config.json':
+                config_files.append(os.path.join(root, f))
+
+    return config_files
+
+
+def main(path):
+    files = find_config_files(path)
+    outfile = open(os.path.join(os.getcwd(), 'config-reference.md'), 'w')
+    outfile.write("# DC/OS Universe Package Configuration Reference\n\n")
+
+    for f in files:
+        with open(f, 'r') as config:
+            package_name = f.split('/')[-3]
+            package_version = f.split('/')[-2]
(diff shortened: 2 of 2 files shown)


rji:
For example, the latest version of the Jenkins package produces the following:

## jenkins version 9

### storage configuration properties

| Property | Type | Description | Default Value |
|----------|------|-------------|---------------|
| host-volume | string | The location of a volume on the host to be used for persisting Jenkins configuration and build data. The final location will be derived from this value plus the name set in `name` (e.g. `/mnt/host_volume/jenkins`). Note that this path must be the same on all DCOS agents. | `/tmp` |
| pinned-hostname | string | An optional DCOS agent hostname to run this Jenkins instance on (e.g. 10.0.0.1). | *No default.* |

### networking configuration properties

| Property | Type | Description | Default Value |
|----------|------|-------------|---------------|
| root-url | string | The web address of the Jenkins installation (e.g. http://jenkins.example.com). This is used by Jenkins to refer to itself when creating links in emails, chat notifications, etc. | *No default.* |
| known-hosts | string | A space-separated list of hosts used to populate the SSH known hosts file on the Jenkins master. | `github.com` |
| virtual-host | string | The virtual host address to configure for integration with Marathon-lb. | *No default.* |
| https-redirect | boolean | Whether Marathon-lb should redirect HTTP traffic to HTTPS. This requires 'virtual-host' to be set. By default, this is false. | `False` |

### service configuration properties

| Property | Type | Description | Default Value |
|----------|------|-------------|---------------|
| cpus | number | CPU shares to allocate to each Jenkins master. | `1.0` |
| mem | number | Memory (in MB) to allocate to each Jenkins master. | `2048.0` |
| name | string | The name of the service to display in the DC/OS dashboard. | `jenkins` |

### advanced configuration properties

| Property | Type | Description | Default Value |
|----------|------|-------------|---------------|
| jvm-opts | string | Optional arguments to pass to the JVM. | `-Xms1024m -Xmx1024m` |
| role | string | The accepted resource roles (e.g. slave_public). By default, this will deploy to any agents with the * role. | `*` |
| docker-credentials-uri | string | An optional URI to be fetched and extracted that contains docker credentials (e.g. file:///etc/docker/docker.tar.gz). | *No default.* |
| docker-image | string | The Docker image to use for the Jenkins service. By default, this package will use the Velocity image in the Mesosphere organization on Docker Hub, which you must be authenticated against. Otherwise, specify the host, image, and tag for the Velocity image on your private Docker Registry. | *No default.* |
| mesos-master | string | URL of this cluster's Mesos master. | `zk://leader.mesos:2181/mesos` |

ssk2:
Very very cool!

rji:
Saw two +1's here, going to hit the button.

PR #591, opened 2016-07-27
[do-not-merge] Artifactory 1.7

ssk2:
Generate a version of Artifactory that is installable on a 1.7 cluster.


jainishshah17:
@ssk2 : Can you add changes we made Yesterday.

PR #609, opened 2016-08-01
Default to "hdfs-role"/"hdfs-principal"

nickbp:
To align with defaults in other frameworks


nickbp:
Just hand-checked the fix on a live cluster, :+1:

PR #610, opened 2016-08-01
Fix default role for confluent-kafka

gabrielhartmann:


gabrielhartmann:
@nickbp:  Tested on 1.8 cluster.
screen shot 2016-08-01 at 11 50 51 am

PR #611, opened 2016-08-01
Buggy url - test

mohitsoni:


ssk2:
@mohitsoni: is this something we can close now?

PR #619, opened 2016-08-03
Pass role to Mesos plugin in Jenkins package

titosand:


repo/packages/J/jenkins/10/config.json
```diff
@@ -63,6 +63,22 @@
                 }
             }
         },
+         "roles": {
+              "description": "Role configuration properties for Jenkins on DC/OS.",
+              "type": "object",
+              "properties": {
+                "jenkins-master-role": {
+                    "description": "The accepted resource roles (e.g. slave_public) that the Jenkins master itself runs as using Marathon. By default, this will deploy to any agents with the * role.",
```
colin-msphere: I would change or remove the example, unless we want users to run masters on nodes with `slave_public`. I could see this example being interpreted as a suggested default.

ssk2: Let's move this example to the end, after the statement about default. Rephrase to something like `For example, to deploy to a public DC/OS agent, set this to 'slave_public'.`

repo/packages/J/jenkins/10/config.json
```diff
@@ -63,6 +63,22 @@
...
+                    "type": "string",
+                    "default": "*"
+                },
+                  "jenkins-agent-role": {
+                      "description": "The role passed to the internal Jenkins configuration that denotes which resources can be launched, so Jenkins agents can then only ever schedule onto nodes with the specified role. By default, this also uses the * role",
```
ssk2: `schedule` -> `scheduled`

PR #623, opened 2016-08-05
more marathon-lb fixes

alberts:


vespian:
@alberts  Shouldn't we release a new package version (->10) ? Not sure if users will be able to catch it.

PR #633, opened 2016-08-12
Add arangodb3-1.0.4

m0ppers:
DC/OS compat fixes. Principal properly set for disc reservations (mesos 1.0)
DC/OS labels properly set for UI service discovery

ping @joerg84 :

that's what it should look like I suppose?


joerg84:
LGTM, tested on Enterprise

PR #638, opened 2016-08-15
Removed quotes from numbers to fix config validation with Confluent C3

keithchambers:
Removed quotes from numbers to fix config validation with Confluent Control Center.

@dbtucker @gabrielhartmann


dbtucker:
LGTM

PR #671, opened 2016-08-29
Adding Microsoft OMS Docker solution.

amitsara:
Please look at https://github.com/microsoft/Oms-docker for more details.
I still need to work on icons and solution description, that I will do in next commits.


amitsara:
Will send another request after fixing package file

PR #674, opened 2016-08-30
Add support for Confluent Platform 3.0.1.

dbtucker:
The confluent-kafka package was updated to reference the updated CP 3.0.1 tarball.   Icons were updated to core Kafka.

The other confluent-* packages were updated to reference the 3.0.1 docker images and leverage the improved naming service within the deployment (removing the need for explicit VIPs)


repo/packages/C/confluent-connect/2/resource.json
```diff
@@ -0,0 +1,14 @@
+{
+  "assets": {
+    "container": {
+      "docker": {
+        "image": "dbtucker/cp-kafka-connect"
```
gabrielhartmann: Did you want an explicit tag on this image?  This is going to default to `latest` I think.

repo/packages/C/confluent-kafka/8/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "packagingVersion": "3.0",
+  "name": "confluent-kafka",
+  "version": "1.1.12-3.0.2",
```
gabrielhartmann: This should be `1.1.12-3.0.1` correct?

repo/packages/C/confluent-schema-registry/2/marathon.json.mustache
```diff
@@ -0,0 +1,55 @@
...
+    "type": "DOCKER",
+    "docker": {
+      "image": "{{resource.assets.container.docker.image}}:{{registry.docker-tag}}",
+	  "forcePullImage": true,
+      "network": "BRIDGE",    
+      "portMappings": [ {
+          "containerPort": 8081,
+          "hostPort": 0,
+          "protocol": "tcp",
+    	  "labels": {
```
gabrielhartmann: There's a <tab> in the whitespace here.

dbtucker: Removed all tabs.

repo/packages/C/confluent-connect/2/resource.json
dbtucker: For now, the tag is a separate value.   The image specification in marathon.json.mustache is built from this value and then the docker-tag.

keithchambers: I recommend sticking with an explicit version of the Docker image in the resource.json file vs. making it a variable.

There are a couple reasons why:
* DC/OS UI shows all the versions installed. This won't be accurate anymore.
* There is a tool that allows on-premise DC/OS users to create a local Universe mirror. We need to understand if there are implications if package assets become user configurable.

/cc @pyronicide for his thoughts.

pyronicide: +1 we really need the image to be a single variable (because of downstream tooling).

dbtucker:
Per Keith's suggestion: removed docker-tag option so that images are now specified completely in resource.json.

PR #684, opened 2016-09-07
Add a preInstallNotes for Cassandra Service

orsenthil:
preInstallNotes gives the minimum configuration required.

### DIFF
--- a/repo/packages/C/cassandra/15/package.json
+++ b/repo/packages/C/cassandra/15/package.json
@@ -5,6 +5,7 @@
   "minDcosReleaseVersion": "1.7",
   "name": "cassandra",
   "packagingVersion": "3.0",
+  "preInstallNotes": "DC/OS Cassandra Service default configuration requires 3 nodes each with 0.3 CPU shares, 1184MB of memory and 272MB of disk.",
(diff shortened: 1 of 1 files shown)


repo/packages/C/cassandra/15/package.json
```diff
@@ -5,6 +5,7 @@
   "minDcosReleaseVersion": "1.7",
   "name": "cassandra",
   "packagingVersion": "3.0",
+  "preInstallNotes": "DC/OS Cassandra Service default configuration requires 3 nodes each with 0.3 CPU shares, 1184MB of memory and 272MB of disk.",
```
mohitsoni: I think we should update the message to following:
```
DC/OS Cassandra Service default configuration requires 3 nodes each with 1.5 CPU shares, 5376MB of memory and 11264MB of disk for running Cassandra Nodes. And, 1 node with 0.5 CPU shares, 2048MB of memory for running the service scheduler.
```

orsenthil: I can change this, but I am not the expert in verifying the accuracy of this statement. I will leave it you / marathon team for the accuracy.

(Pushing an update with the suggested change).

PR #692, opened 2016-09-13
update gen_universe.py to create 1.9 + 1.10 versioned universe jsons

tamarrow:
Modularized code so adding new versions will be much simpler


BenWhitehead:
Manually tested nginx config changes locally after building universe with updated script.

tamarrow:
nginx changes lgtm. @BenWhitehead merge when tests pass?

PR #696, opened 2016-09-15
Gestalt version 0.2.0

cgbaker:


pyronicide:
Please move this to a `1` directory. This is a major change and should be treated as a brand new version.

cgbaker:
@pyronicide : i'm struggling to add a `minDcosReleaseVersion` to the package.json. every time i upgrade `packagingVersion` to `3.0` in the package.json, the result of the build is that the package marathon mustache file doesn't seem to get processed. there's no indication as to why that is in the output:
```
Validating gestalt-framework...
	Validating revision 0...
		package.json:	OK
		config.json:	OK
		marathon.json.mustache:	OK
		resource.json:	OK
	Validating revision 1...
		package.json:	OK
		config.json:	OK
		resource.json:	OK
Validating gitlab...
	Validating revision 0...
		package.json:	OK
```

pyronicide:
@cgbaker are you adding all the keys? See:

https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/C/cassandra/15/package.json

repo/packages/G/gestalt-framework/1/resource.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images": {
+    "icon-large": "https://s3.amazonaws.com/gfi.misc/Gestalt-Icon-Large.png",
+    "icon-medium": "https://s3.amazonaws.com/gfi.misc/Gestalt-Icon-Medium.png",
+    "icon-small": "https://s3.amazonaws.com/gfi.misc/Gestalt-Icon-Small.png"
+  },
+  "assets": {
+    "container": {
+      "docker": {
```
pyronicide: You need to include the docker here (otherwise on-prem folks won't be able to use this).

cgbaker: urgh, tough call. the framework comprises of multiple docker images, versioned ensemble. i can register them all in the `resource.json` file to facilitate on-prem docker registries. however, the universe documentation requests that these also use the version ID for the docker images, which means that we're forced to do a whole new release for a hotfix, instead of just moving a tag forward (previously, to avoid having to do a new PR against against mesosphere/universe). 

(moving tags does seem rude, so maybe we should avoid it in general.)

the practice, as evidenced by other packages in universe, seems to be all over the place. most packages don't use the Image ID; a few packages even use inconstant `latest`. we're fine with dropping `latest` and going through PRs to roll out new versions, and even packaging with the image IDs (we would probably rely on our own universe server to support hot-fixes).

i'd appreciate your suggestion. thanks.

pyronicide: We'd like to have a new version for each release (even hotfixes). We're going to have a more automated process here soon to make it way easier to get new versions in and going.

cgbaker: cool, i added all of the framework docker images to the metadata to support the on-prem registries

PR #701, opened 2016-09-20
Add Calico v2 (v0.3.0)

djosborne:


repo/packages/C/calico/2/config.json
```diff
@@ -0,0 +1,245 @@
...
+    "Framework Settings": {
+      "description": "Configuration options for Calico's DC/OS Framework.",
+      "properties": {
+        "framework-name": {
+          "type": "string",
+          "description": "The name of the framework.",
+          "default": "calico-install-framework"
+        },
+        "master": {
+          "default": "zk://master.mesos:2181/mesos",
```
pyronicide: We should probably switch this over to the new zk string.

`zk://zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4:2181,zk-5:2181/mesos`

djosborne: I switched default to `zk://zk-1.zk:2181/calico`, allowing the user to expand it to the other zk's if they're running a multi cluster. Is that sufficient, or should it be the full string?

pyronicide: We do round-robin on the backend depending on the number of zks someone is running (in a single master cluster, they all resolve to the same address). We're doing this because of issues in zk drivers around network partitions and dynamic masters.

djosborne: ok cool. So if I understand correctly, always use the full string, because the "extra" `zk`'s will still resolve to real zk's?

pyronicide: Exactly!

JohnOmernik:
We'd love to see this merged!

pyronicide:
Just waiting on the zk url fix. Once thats in, its ready to merge!

djosborne:
@pyronicide fixed!

PR #716, opened 2016-09-26
Openldap and openldap-admin 

fernandosanchezmunoz:
add "networking" section. Add "external_access" flag to enable VIP on Marathon-LB optionally.
add "storage" section. add "persistence" flag to optional create local persistent volumes.
change to TCP healthcheck.
fix docker versions where possible (some images just offer latest)
add icons
DCOS_PACKAGE_IS_FRAMEWORK=false
bump to version 0.3


repo/packages/O/openldap-admin/2/marathon.json.mustache
```diff
@@ -0,0 +1,52 @@
...
+        "docker": {
+            "image": "{{resource.assets.container.docker.openldap-admin-docker}}",
+            "forcePullImage": false,
+            "network": "BRIDGE",
+            "portMappings": [
+            {
+                "containerPort": 80,
+                "hostPort": 0,
+                {{#networking.external_access}}
+                "servicePort": {{networking.port_external}},
```
pyronicide: You're getting "external" access via. adminrouter. Isn't that sufficient?

repo/packages/O/openldap/2/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "packagingVersion": "2.0",
```
pyronicide: Need a minVersion of at least 1.7 because of the storage.

repo/packages/O/openldap-admin/2/marathon.json.mustache
fernandosanchezmunoz: I'm defaulting to external_access=false so that the VIP won't be created. I can see use cases where:
- AdminRouter is used exclusively for Dan access.
- AdminRouter is bloated with traffic
- The cluster is used for "outside" access for people not exposed at all to DC/OS

... so I thought it'd be a good idea to give the *option* to automatically put it in MLB

repo/packages/O/openldap/2/package.json
fernandosanchezmunoz: can I simply change this or does it have any other implications? TBH I was reusing from other existing packages so if there's anything I can read on what this implies I'll check it out.

fernandosanchezmunoz: done thanks

PR #719, opened 2016-09-27
Mysql + mysql admin

fernandosanchezmunoz:
#- removed the option for bridge/host networking (todo: bring it back and add to other similar packages)
- added persistent storage section and capabilities
- added external connectivity section and capabilities
- added MySQL-admin based on phpmyadmin
- change container path in mounted volume to "var" to avoid issues with using "/"
- MySQL will always create a new database, user and password with default values

### DIFF
--- a/repo/packages/M/mysql/4/package.json
+++ b/repo/packages/M/mysql/4/package.json
@@ -1,0 +1,20 @@
+{
+  "packagingVersion": "3.0",
+  "minDcosReleaseVersion": "1.7",
+  "name": "mysql",
+  "version": "0.2",
+  "scm": "https://github.com/mysql/mysql-server.git",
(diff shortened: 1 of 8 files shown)


repo/packages/M/mysql/4/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "packagingVersion": "3.0",
+  "minDcosReleaseVersion": "1.7",
```
pyronicide: 1.8?

fernandosanchezmunoz: local volumes in 1.7 

https://docs.mesosphere.com/1.7/usage/storage/persistent-volume/

PR #726, opened 2016-09-28
Define ElasticSearch version through a config parameter

albertostratio:
Currently the ElasticSearch framework is installing the latest ElasticSearch docker image so you can't use a previous version of ElasticSearch.

This PR includes the changes to allow the user to define the docker image he wants the executors to use. A new field will be included within the DC/OS Universe advanced installation to allow the user to set a different version of Elasticsearch.

### DIFF
--- a/repo/packages/E/elasticsearch/1/config.json
+++ b/repo/packages/E/elasticsearch/1/config.json
@@ -42,6 +42,11 @@
           "description":"Forces docker to re-pull the image.",
           "type":"boolean"
         },
+        "docker-image":{
+          "default":"elasticsearch:latest",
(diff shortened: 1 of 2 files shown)


triclambert:
Hi @albertostratio, this seems like a pretty substantial change to a package that other people may already be using.  Instead, I think a better approach would be to submit a new package (version 2) with these changes.  This way folks who are may have come to depend on the behavior of version 1 will not be surprised.  Make sense?  

/cc @ryadav88 our new Partner Manager

albertostratio:
Thank you for your comment @triclambert. It makes complete sense. I'll close this PR and will open a new one adding a new version with these changes. Thank you again!

PR #732, opened 2016-09-29
mariadb: update to bitnami/mariadb:10.1.18-r0 release

sameersbn:
- includes OpenSSL security fixes


sameersbn:
updated to `bitnami/mariadb:10.1.18-r0`

PR #740, opened 2016-10-01
mysql improvement set inc. host mode

fernandosanchezmunoz:
* correct install Notes and improve info about configurable parameters
* fix scm links where appropriate
* add screenshots in the admin apps
* add note recommending to use endpoint in marathon-LB vs admin router to remove load from AR when possible
* remove configurable "instances" qty - fix to 1
* add "MARATHON_SINGLE_INSTANCE_APP": "true",
* add "upgradeStrategy" (0/0) for single instance apps
* remove "defaults" comments - defaults are obvious when configuring
* use coherent variable names across apps
* use standard default credentials where appropriate: "admin" "password"
* remove "version" from config.json
* move persistent storage to top of description, host_volume to bottom
* fix descriptions of cpus and mem for homogeneity and clarity
* add HOST mode for performance
* ports are only configurable only in bridge option - most dockers don't listen to $PORT0
* add "requirePorts" if host mode
* add HOSTNAME: UNIQUE if host mode
* change version format to $SOFTWAREVERSION-PACKAGEVERSION
e.g. mysql version 6.5.2, package version 0.2 gives
6.5.2-0.2
* bump package minor version

### DIFF
--- a/repo/packages/M/mysql-admin/1/config.json
+++ b/repo/packages/M/mysql-admin/1/config.json
@@ -1,0 +1,80 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "properties": {
+        "service":{
+            "type":"object",
+            "description": "DC/OS service configuration properties",
+            "properties":{
+                "name" : {
+                    "description":"Name of this service instance.",
+                    "type":"string",
+                    "default":"mysql-admin"
+                }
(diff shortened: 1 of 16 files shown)


fernandosanchezmunoz:
closing as there's a new update

PR #748, opened 2016-10-04
gen-universe: packages should be deep, not shallow copied

tamarrow:
Mutable data structure caused removed of resource.cli property from
versions that accepted it, since earlier versions that DO NOT accept
that property were processed first.


tamarrow:
https://giphy.com/gifs/embarrassed-facepalm-panda-14aUO0Mf7dWDXW

PR #773, opened 2016-10-20
update preInstallNotes for marathon-lb

ryadav88:
Adding a note about setting a Service Account in EE DC/OS


ryadav88:
@pyronicide Let me know if this looks good or needs more change.

PR #781, opened 2016-10-25
Update GitLab to version 8.13.0

twk3:
Update GitLab to 8.13.0, released on Oct 22nd

This update brings improved root_squash support in the case of users setting up NFS on the hostmounts.


twk3:
cc\ @ssk2 

Tested on DC/OS 1.8.6
Tested an upgrade on a single-node pinned instance, and tested install of both CE and EE image

ssk2:
Thanks @twk3!

PR #884, opened 2016-12-20
Ceph dash

fernandosanchezmunoz:
A tiny little dashboard to see how you're DC/OS Ceph cluster is doing today.

Tutorial for this one will be merged immediately in the main Ceph tutorial as a section. Authentication is common to both packages and... hey, who would run a Ceph cluster and NOT want to see how it's doing?

### DIFF
--- a/repo/packages/C/ceph-dash/0/resource.json
+++ b/repo/packages/C/ceph-dash/0/resource.json
@@ -1,0 +1,18 @@
+{
+  "images": {
+    "icon-small": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-small.png",
+    "icon-medium": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-medium.png",
+    "icon-large": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-large.png",
+    "screenshots": [
+      "https://raw.githubusercontent.com/crapworks/ceph-dash/master/screenshots/ceph-dash.png",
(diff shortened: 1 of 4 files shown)


fernandosanchezmunoz:
can you please merge this and the new Ceph so that I can do the example over the merged packages instead of handling two different branches?

repo/packages/C/ceph-dash/0/resource.json
```diff
@@ -0,0 +1,18 @@
...
+    "icon-large": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-large.png",
+    "screenshots": [
+      "https://raw.githubusercontent.com/crapworks/ceph-dash/master/screenshots/ceph-dash.png",
+      "https://raw.githubusercontent.com/crapworks/ceph-dash/master/screenshots/ceph-dash-graphite.png"
+    ]
+  },
+  "assets": {
+    "container": {
+      "docker": {
+        "ceph-dash-docker": "fernandosanchez/ceph-dash:0.1"
```
ryadav88: should be pulled from `mesosphere` org so we can get usage of packages created by Mesosphere

ryadav88:
> can you please merge this and the new Ceph so that I can do the example over the merged packages instead of handling two different branches?

Ok for this one but the point of asking for examples is that the reviewer can test the package based on the `example`. This will change when we have tests in place.

repo/packages/C/ceph-dash/0/resource.json
fernandosanchezmunoz: happy to host it at the mesosphere account, let's work to see how we can get that done.

PR #885, opened 2016-12-21
ceph version 0.2

fernandosanchezmunoz:
Default to no network authentication (0.0.0.0/0) so that the package installs without manual configuration required.
Also, add lighter docker image.

### DIFF
--- a/repo/packages/C/ceph/1/resource.json
+++ b/repo/packages/C/ceph/1/resource.json
@@ -1,0 +1,14 @@
+{
+  "images": {
+    "icon-small": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-small.png",
+    "icon-medium": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-medium.png",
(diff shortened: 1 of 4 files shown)


fernandosanchezmunoz:
also changed to lighter ceph-on-mesos:0.1 docker image.

repo/packages/C/ceph/1/resource.json
```diff
@@ -0,0 +1,14 @@
+{
+  "images": {
+    "icon-small": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-small.png",
+    "icon-medium": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-medium.png",
+    "icon-large": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-large.png"
+  },
+  "assets": {
+    "container": {
+      "docker": {
+        "ceph-docker": "fernandosanchez/ceph-on-mesos:0.1"
```
ryadav88: should be pulled from `mesosphere` org so we can get usage of packages created by Mesosphere

fernandosanchezmunoz: happy to host it at the mesosphere account, let's work to see how we can get that done.

PR #893, opened 2016-12-31
Bump Docker registry version in the `local-universe.py` script

liorze:
This PR bumps the Docker registry version that is started by the `local-universe.py` script to 2.4.1 in order to match the one that is required by the base image (2.4.1). This way we don't have to download another registry image, which might save a few minutes if that version is not already cached in the machine.

### DIFF
--- a/scripts/local-universe.py
+++ b/scripts/local-universe.py
@@ -188,7 +188,7 @@ def run_docker_registry(volume_path):
     print('Start docker registry.')
     command = [ 'docker', 'run', '-d', '-p', '5000:5000', '--name',
         'registry', '-v', '{}:/var/lib/registry'.format(volume_path),
-        'registry:2.4.0']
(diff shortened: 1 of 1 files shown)


jsancio:
LGTM.

tobilg:
I see the same behavior, `sudo make base` downloads `2.4.1`, `sudo make local-universe` downloads `2.4.0`

ryadav88:
@pyronicide  can you review and merge this?

cruhland:
I tested that generation of `local-universe.tar.gz` works correctly. @liorze did you confirm that the repo works correctly when added to DC/OS?

cruhland:
Got around to testing the result in a DC/OS cluster. I was able to load a local universe and launch the zeppelin package successfully. :ship: it @ryadav88!

PR #910, opened 2017-01-13
dcos-enterprise-cli: push updates to windows CLI

tamarrow:


repo/packages/D/dcos-enterprise-cli/10/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packagingVersion": "3.0",
+  "name": "dcos-enterprise-cli",
+  "version": "1.0.7",
```
vespian: Do we need to bump the rev of the package ?

I think it does not matter for Linux and Mac binaries, but fixed Windows version needs to bumped hence bumping all of them..

tamarrow: The binary itself isn't different, its the zip folder format. In this case I'm going to keep the version because I think the same CLI code should be the same version.

vespian: 👍

PR #913, opened 2017-01-18
Added artifactory 5.x-SNAPSHOT

jainishshah17:
to be used for webinar next week.


ryadav88:
@jainishshah17 just talked to @ssk2 
Since this will only be used for the webinar, we should build the package using local universe. 
Closing this PR. Feel free to comment if you think it should be reopened.

PR #928, opened 2017-01-30
tunnel-cli: Bump to version 1.0.0

nlsun:
This includes a update to the version of pyinstaller used. This fixes a bug where the environment for python subprocess calls were dirty and caused openvpn to fail.

### DIFF
--- a/repo/packages/T/tunnel-cli/5/package.json
+++ b/repo/packages/T/tunnel-cli/5/package.json
@@ -1,0 +1,9 @@
+{
+  "packagingVersion": "3.0",
+  "name": "tunnel-cli",
+  "version": "0.3.3",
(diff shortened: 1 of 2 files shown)


repo/packages/T/tunnel-cli/5/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "packagingVersion": "3.0",
+  "name": "tunnel-cli",
+  "version": "1.0.0",
```
ryadav88: is this really a version bump from 0.3.3 to 1.0.0?

nlsun: yup. I figured that now is as a good a time as ever to bump to 1.0.0.

PR #941, opened 2017-02-06
Fixed zeppelin documentation links.

joerg84:


joerg84:
P.S. In-place change as it only touches description.

PR #942, opened 2017-02-06
Increased grace period for MoM.

unterstein:
When migrating, marathon will not respond to health checks until all applications are migrated. When migrating a lot of applications, marathon will -in some cases- not be able to finish migration within the grace period. Therefore this period should be increased.

### DIFF
--- a/repo/packages/M/marathon/0/marathon.json.mustache
+++ b/repo/packages/M/marathon/0/marathon.json.mustache
@@ -14,7 +14,7 @@
   "uris": {{marathon.uris}},
   "healthChecks": [
     {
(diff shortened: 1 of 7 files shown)


jdef:
@unterstein is there a plan to change this marathon behavior? if marathon is healthy enough to elect a leader and begin migration .. why is it not healthy enough to respond to health checks? seems like "i'm busy because i'm going migration" is some kind of readiness thing vs. a health thing

unterstein:
Yes. for 1.3 and newer one we should use readiness checks instead of health checks. @meichstedt already worked on this issue

PR #963, opened 2017-02-14
Add v4Package Schema

larioj:
We are adding updates to packages. As part of this effort we are adding a new package specification which encodes upgrade information. Specifically we are adding a field that shows for a given package `A`, what other packages `upgrades=[B, C, D]` are able to upgrade to `A`.

### DIFF
--- a/repo/meta/schema/package-schema.json
+++ b/repo/meta/schema/package-schema.json
@@ -295,8 +295,7 @@
         "version",
         "maintainer",
         "description",
-        "tags",
-        "minDcosReleaseVersion"
+        "tags"
       ],
       "additionalProperties": false
     }
--- a/repo/meta/schema/package-schema.json
+++ b/repo/meta/schema/package-schema.json
@@ -198,7 +198,7 @@
       "properties": {
         "packagingVersion": {
           "type": "string",
-          "enum": ["3.0"]
+          "enum": ["4.0"]
         },
         "name": {
           "type": "string"
@@ -294,7 +294,7 @@
   },
 
   "type": "object",
-  "anyOf": [
+  "oneOf": [
     { "$ref": "#/definitions/v20Package" },
     { "$ref": "#/definitions/v30Package" },
     { "$ref": "#/definitions/v40Package" }
--- a/repo/meta/schema/package-schema.json
+++ b/repo/meta/schema/package-schema.json
@@ -223,12 +223,19 @@
           "default": false,
           "description": "True if this package installs a new Mesos framework."
         },
-        "upgrades": {
+        "upgradesFrom": {
           "type": "array",
           "items": {
             "type": "string"
           },
-          "description": "List of versions from which this package can upgrade from. If the property is missing or null it means that it can upgrade from any package. If the property is the empty list then it means that it can’t upgrade from any package."
+          "description": "Versions this package can upgrade from."
+        },
+        "downgradesTo": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "Versions this package can downgrade to."
         },
         "preInstallNotes": {
           "type": "string",
@@ -286,7 +293,9 @@
         "version",
         "maintainer",
         "description",
-        "tags"
+        "tags",
+        "upgradesFrom",
+        "downgradesTo"
       ],
       "additionalProperties": false
     }
--- a/repo/meta/schema/package-schema.json
+++ b/repo/meta/schema/package-schema.json
@@ -114,7 +114,8 @@
           "type": "string"
         },
         "version": {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[-a-zA-Z0-9.]+$"
         },
         "scm": {
           "type": "string"
@@ -204,7 +205,8 @@
           "type": "string"
         },
         "version": {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[-a-zA-Z0-9.]+$"
         },
         "scm": {
           "type": "string"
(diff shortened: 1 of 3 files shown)


repo/meta/schema/package-schema.json
```diff
@@ -192,14 +192,112 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "True if this package installs a new Mesos framework."
+        },
+        "upgrades": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of versions from which this package can upgrade from. If the property is missing or null it means that it can upgrade from any package. If the property is the empty list then it means that it can’t upgrade from any package."
```
cruhland: Some suggestions:
1. "List of versions that can upgrade to this package."
1. "If the property is missing or null, any version can upgrade to this package."
1. "If the property is the empty array, no version can upgrade to this package."

cruhland:
Also, I assume we're not creating a `version-4.x` branch of the repo this time, since we now have the schema versions?

repo/meta/schema/package-schema.json
```diff
@@ -114,7 +114,8 @@
           "type": "string"
         },
         "version": {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[-a-zA-Z0-9.]+$"
```
cruhland: Did you check that this is compatible with all versions currently in Universe? Or is this just the characters allowed by semver?

repo/meta/schema/package-schema.json
```diff
@@ -192,14 +193,123 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "True if this package installs a new Mesos framework."
+        },
+        "upgradesFrom": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of versions that can upgrade to this package. If the property is a list containing '*', any version can upgrade to this package. If the property is the empty list, no version can upgrade to this package."
```
cruhland: What happens if the list has both version numbers and `*`?

scripts/validate-packages.py
```diff
@@ -57,6 +57,15 @@ def _validate_revision(given_package, revision, path):
...
     packaging_version = package_json.get("packagingVersion", "2.0")
 
+    # validate upgrades version
+    if packaging_version == "4.0":
+        min_dcos_version = package_json.get("minDcosReleaseVersion").split('.')
+        major = int(min_dcos_version[0])
+        minor = int(min_dcos_version[1])
+
+        if major < 1 or (major == 1 and minor < 10):
+            sys.exit("\tERROR\n\nminDcosReleaseVersion must be greater than or equal to 1.10")
```
cruhland: Looks like the other checks in this file use `eprint()`?

scripts/validate-packages.py
```diff
@@ -57,6 +57,15 @@ def _validate_revision(given_package, revision, path):
 
     packaging_version = package_json.get("packagingVersion", "2.0")
 
+    # validate upgrades version
+    if packaging_version == "4.0":
+        min_dcos_version = package_json.get("minDcosReleaseVersion").split('.')
+        major = int(min_dcos_version[0])
+        minor = int(min_dcos_version[1])
+
+        if major < 1 or (major == 1 and minor < 10):
```
cruhland: Let's follow the existing conventions for comparing release versions in this code: https://github.com/larioj/universe/blob/7465bacd4c151afde330bbfffc70832dd7854f46/scripts/validate-packages.py#L109

cruhland:
@larioj overall looks good, had a few comments.

repo/meta/schema/package-schema.json
larioj: I checked that this is compatible with all current v3Package versions. Also the fact that the CI passes proves this.

repo/meta/schema/package-schema.json
larioj: ` If the property is a list containing '*', this package can downgrade to any version.` Any list containing star is equivalent to ['*'].

scripts/validate-packages.py
larioj: The checks that say OK, or some Info use eprint(). The checks that fail the verification use sys.exit()

scripts/validate-packages.py
larioj: Thanks. That's a lot easier.

repo/meta/schema/package-schema.json
jsancio: How about `... If the property is a list containing the string "*", any...`?

repo/meta/schema/package-schema.json
```diff
@@ -192,14 +193,123 @@
...
+      },
+      "required": [
+        "packagingVersion",
+        "name",
+        "version",
+        "maintainer",
+        "description",
+        "tags",
+        "upgradesFrom",
+        "downgradesTo",
```
jsancio: Okay, are we going to convert these packages to an object/package without these two properties (upgradesFrom, downgradTo) if the values are the empty array?

scripts/validate-packages.py
```diff
@@ -57,6 +57,11 @@ def _validate_revision(given_package, revision, path):
 
     packaging_version = package_json.get("packagingVersion", "2.0")
 
+    # validate upgrades version
+    min_dcos_release_version = package_json.get("minDcosReleaseVersion")
+    if packaging_version == "4.0" and LooseVersion(min_dcos_release_version) < LooseVersion("1.10"):
+        sys.exit("\tERROR\n\nminDcosReleaseVersion must be greater than or equal to 1.10")
```
jsancio: Can we do better? Can we convert these packages to version 3.0 if both upgradeFrom and downgradeTo are the empty list? What do you think?

larioj: Absolutely, I made them optional. If upgradesFrom, and downgradesTo are null, then they are equivalent to the empty list.

larioj: Converting V4 to V3 makes sense under these semantics.

repo/meta/schema/package-schema.json
larioj: Yes

scripts/validate-packages.py
jsancio: @larioj Don't we need to update this check after your change?

PR #965, opened 2017-02-15
Add package instana-agent to DC/OS universe.

doertedev:
This PR includes a package for the instana-agent for SaaS and On-premise customers.


Rakefile
```diff
@@ -0,0 +1,34 @@
+REGISTRY = 'registry-agent.instana.io'
```
joerg84: @doertedev can we drop the Rakefile again?

PR #966, opened 2017-02-15
Add V40Package to repo schema

larioj:
Update repository schema to accept `V40Package`s. We are adding updates to packages. As part of this effort we are adding a new package specification which encodes upgrade information. Specifically we are adding a field that shows for a given package A, what other packages upgrades=[B, C, D] are able to upgrade to A.

### DIFF
--- a/repo/meta/schema/v3-repo-schema.json
+++ b/repo/meta/schema/v3-repo-schema.json
@@ -482,6 +482,121 @@
         "tags"
       ],
       "additionalProperties": false
+    },
+
+    "v40Package": {
+      "properties": {
+        "packagingVersion": {
+          "type": "string",
+          "enum": ["4.0"]
+        },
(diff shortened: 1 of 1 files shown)


cruhland:
@larioj what's the difference between this PR and #963? Why are there two schemas?

larioj:
@cruhland There are two schemas. This is the repo schema, i.e. the schema for the thing that universe serves cosmos. And there is the packaging schema. Which is the schema of the packages. The repo schema contains packages, so it has the package schema embedded.

joerg84:
@larioj @cruhland Is this ready to be merged or do we need -as of last comment- more changes?

larioj:
@joerg84 We had a meeting that changed the scope of things. So this PR will have to be updated to reflect that. I will remove the WIP label when it's ready.

PR #970, opened 2017-02-16
tunnel-cli: Bump to 1.0.2

nlsun:
Config parsing fixes.


joerg84:
LGTM

joerg84:
@nlsun 
a) It would be great to have an example for this at https://github.com/dcos/examples
b) Just curious: This does not work under windows? at least no binary?

nlsun:
The reason why we labelled it as not working under windows is because there is a lot of shelling out occurring, which isn't known to work particularly well under windows.

PR #979, opened 2017-02-23
Add Flink jobmanager's log file to sandbox

skonto:
In some cases Flink Jobmanager will fail fast, before we even have access to the its ui. It is is good to have also the file in sandbox to download it easily from the dc/os ui.

### DIFF
--- a/repo/packages/F/flink/0/marathon.json.mustache
+++ b/repo/packages/F/flink/0/marathon.json.mustache
@@ -4,7 +4,7 @@
   "mem": {{app-master.memory}},
   "env": {
     "DCOS_SERVICE_NAME": "{{service.name}}",
-    "FLINK_LOG_DIR": "/mnt/mesos/sandbox"
+    "FLINK_LOG_DIR": "$MESOS_SANDBOX"
   },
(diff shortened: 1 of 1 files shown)


skonto:
@joerg84 pls review. 
Note: I was looking at $MESOS_SANDBOX so I have no hardcoded value there, but I think it is filled with a value when the docker command is launched (it is availble in the container's env) not previously right?

repo/packages/F/flink/0/marathon.json.mustache
```diff
@@ -3,7 +3,8 @@
   "cpus": {{app-master.cpus}},
   "mem": {{app-master.memory}},
   "env": {
-    "DCOS_SERVICE_NAME": "{{service.name}}"
+    "DCOS_SERVICE_NAME": "{{service.name}}",
+    "FLINK_LOG_DIR": "/mnt/mesos/sandbox"
```
joerg84: Should we $MESOS_SANDBOX instead here (https://mesosphere.github.io/marathon/docs/task-environment-vars.html)?

skonto: That's my question if you check the comment above I saw that it is used inside the bash command in the docker not sure if it would work there. I will change it.

skonto:
Btw I noticed that also we get the Taskmanager's log file in the sandbox.

joerg84:
LGTM

PR #984, opened 2017-02-27
Update resource URL's

iamtew:
Pointing the URL's to the raw.githubusercontent.com instead as it has
HTTPS enabled. Doing this will prevent browsers from displaying
warnings about accessing mixed content when people are browsing the
Universe through the DC/OS UI and have HTTPS enabled.


iamtew:
Any update on this? Is changes still required?

PR #988, opened 2017-03-01
Add v4 package conversions and fix bug

larioj:


scripts/gen-universe.py
```diff
@@ -98,13 +98,15 @@ def render_json_by_version(outdir, packages, version):
     :rtype: None
     """
 
+    packages = list(
+        filter(lambda package: filter_by_version(package, version), packages))
```
cruhland: It's more idiomatic in Python to use a list comprehension:

```python
packages = [package for package in packages if filter_by_version(package, version)]
```

scripts/gen-universe.py
```diff
@@ -98,13 +98,15 @@ def render_json_by_version(outdir, packages, version):
...
+
+    if LooseVersion(version) < LooseVersion('1.10'):
+        packages = list(map(downgrade_package_to_v3, packages))
```
cruhland: Similarly, you can use a list comprehension here:

```python
packages = [downgrade_package_to_v3(package) for package in packages]
```

scripts/gen-universe.py
```diff
@@ -409,17 +411,12 @@ def write_package_in_zip(zip_file, path, package):
     :rtype: None
     """
 
-    package = copy.deepcopy(package)
+    package = downgrade_package_to_v2(package)
+
     package.pop('releaseVersion')
```
cruhland: Why not move this into `downgrade_package_to_v2()`?

scripts/gen-universe.py
```diff
@@ -409,17 +411,12 @@ def write_package_in_zip(zip_file, path, package):
...
-    package.pop('minDcosReleaseVersion', None)
-    package['packagingVersion'] = "2.0"
 
     resource = package.pop('resource', None)
```
cruhland: Won't `downgrade_package_to_v2()` have already removed this field, because it calls `v3_to_v2_package()`?

scripts/gen-universe.py
```diff
@@ -499,5 +496,78 @@ def create_index_entry(packages):
...
+    package.pop('minDcosReleaseVersion', None)
+    package['packagingVersion'] = "2.0"
+
+    resource = package.pop('resource', None)
+    if resource:
+        cli = resource.pop('cli', None)
+        if cli and 'command' not in package:
+            print(('WARNING: Removing binary CLI from ({}, {}) without a '
+                  'Python CLI').format(package['name'], package['version']))
+        package['resource'] = resource
```
cruhland: Instead of removing the resource object and adding it back, why not just use `.get()` and modify the resource object directly?

scripts/gen-universe.py
```diff
@@ -499,5 +496,78 @@ def create_index_entry(packages):
...
+    :return: a v3 package
+    :rtype: dict
+    """
+    package = copy.deepcopy(v4_package)
+    package.pop('upgradesFrom', None)
+    package.pop('downgradesTo', None)
+    return package
+
+
+def downgrade_package_to_v2(package):
```
cruhland: Instead of explicit downgrade functions, what if you had something like this?

```python
convert_from = {'4.0': v4_to_v3_package, '3.0': v3_to_v2_package}

def downgrade_package_to(package, target_version):
    # TODO
```

Then implement `downgrade_package_to()` so that it repeatedly looks up the package's version in `convert_from` and applies the associated function, until the package reaches `target_version`.

scripts/gen-universe.py
cruhland: Never mind, I see what's happening. But I left another comment about it.

scripts/gen-universe.py
larioj: The release version is not part of the package schema. Removing releaseVersion is just part of the conversion to the zip folder structure.

scripts/gen-universe.py
larioj: KISS. The code you are proposing is a bit more complicated that just unrolling the loop. I agree this should be done. But let's leave it for the next time we add a package version.

scripts/gen-universe.py
larioj: Good catch :)

scripts/gen-universe.py
jsancio: Let's call this `convert_...`.

scripts/gen-universe.py
```diff
@@ -499,5 +495,77 @@ def create_index_entry(packages):
...
+    packaging_version = package.get("packagingVersion")
+    if packaging_version == "2.0":
+        return copy.deepcopy(package)
+    elif packaging_version == "3.0":
+        return v3_to_v2_package(package)
+    else:
+        return v3_to_v2_package(v4_to_v3_package(package))
+
+
+def downgrade_package_to_v3(package):
```
jsancio: Let's call this `convert_package_to_v3`

larioj: That name is a little misleading since the function does not convert v2 packages to v3 packages.

jsancio: I see. What you want is more like `convert_to_dcos_1_8`, no?

scripts/gen-universe.py
jsancio: `convert_to_at_least_v2`

scripts/gen-universe.py
jsancio: Actually, `convert_to_at_most_v3`, no?

larioj: convert_to_v3_if_greater_than_v3?

larioj: It's a hard name. I thought downgrade was appropriate since it conveys the conversion only happens from larger versions to smaller ones.

jsancio: Sounds good!

larioj: Discussed off line. Will be leaving names as is.

scripts/gen-universe.py
larioj: Discussed off line. Will be leaving names as is.

scripts/gen-universe.py
cruhland: :+1:

PR #1012, opened 2017-03-10
Make server return correct content type for DC/OS 1.10

larioj:
This PR updates the nginx conf for the universe-server. The universe-server is a docker image that we generate on every PR, so that package developers can have a test universe with their package in it.

This change is necessary so that the universe server returns the correct universe version for DCOS clusters of version 1.10. In 1.10 we have created a new packaging format, which means a new universe version. (v4)

### DIFF
--- a/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
+++ b/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
@@ -69,6 +69,7 @@ server {
   }
 
   location = /v3 {
+    internal;
     types {
       "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3" json;
     }
@@ -76,6 +77,7 @@ server {
   }
 
   location = /v4 {
+    internal;
     types {
       "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4" json;
     }
--- a/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
+++ b/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
@@ -12,6 +12,12 @@ server {
 
   if ($http_accept = "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3") {
     set $serve_json true;
+    set $universe_version v3;
+  }
+
+  if ($http_accept = "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4") {
+    set $serve_json true;
+    set $universe_version v4;
   }
 
   if ($http_user_agent ~ ".*dcos\/1\.8.*") {
@@ -48,15 +54,32 @@ server {
 
   location = /repo {
     types {
-      "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3" json;
       application/zip zip;
     }
 
-    if ($serve_json = true) {
-      rewrite ^/repo$ /repo-up-to-$dcos_release_version.json break;
+    if ($universe_version = v3) {
+      rewrite ^/repo$ /v3;
+    }
+
+    if ($universe_version = v4) {
+      rewrite ^/repo$ /v4;
     }
 
     rewrite ^/repo$ /repo-up-to-$dcos_release_version.zip break;
+  }
+
+  location = /v3 {
+    types {
+      "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3" json;
+    }
+    rewrite ^/v3$ /repo-up-to-$dcos_release_version.json break;
+  }
+
+  location = /v4 {
+    types {
+      "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4" json;
+    }
+    rewrite ^/v4$ /repo-up-to-$dcos_release_version.json break;
   }
 
 }
--- a/docker/local-universe/default.conf
+++ b/docker/local-universe/default.conf
@@ -14,6 +14,12 @@ server {
 
   if ($http_accept = "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3") {
     set $serve_json true;
+    set $universe_version v3;
+  }
+
+  if ($http_accept = "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4") {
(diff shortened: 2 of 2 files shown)


cruhland:
@larioj have you tested this?

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -72,13 +72,25 @@ server {
...
     rewrite ^/v3$ /repo-up-to-$dcos_release_version.json break;
   }
 
   location = /v4 {
     types {
       "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4" json;
     }
+    
+    if ($dcos_release_version != "1.10") {
+      rewrite ^/v4$ /repo-empty-v3.json break; # empty v3 = empty v4
```
cruhland: Hmm, this seems a bit hacky. Can we generate an empty v4 file?

larioj: Yeah, we can. I can do it as part of this PR.

larioj: On second thought it's actually a lot of work. I think this whole subsystem needs to be simplified. We have multiple nginx conf files that are all similar, but not the same, scattered throughout repositories.

larioj: The correct way to do this is to have the gen-universe.json generate all possible combinations with names `repo-up-to-${dcos-release-version}-${universe-version}.${format}. Where format is either `zip` or `json`. Then just do rewrites to the correct file based on the headers the users give us. It would make the nginx conf files a lot simpler, and we could use the same one in all places.

larioj: It would also make it easier to encode content-types, and (depending on how we do the rewrite logic) we could make it generic enough so that we may not need to update with every dcos-release version.

larioj: But for now this is fine since this change only applies to universe server and not the production universe.

larioj: Also the above is a big change. And I don't know which parts are user facing.

cruhland: You're right, but it's likely not going to be worth doing everything you suggest, since we're planning on replacing it with universe-service soon. Let's just generate the empty v4 file for now.

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -48,15 +54,44 @@ server {
...
   }
 
+  location = /v3 {
+    types {
+      "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3" json;
+    }
+    
+    if ($dcos_release_version = "1.10") {
+      rewrite ^/v3$ /repo-empty-v3.json break;
+    }
```
cruhland: Ok, now that I see how this works, I changed my mind. Let's not have a special case for 1.10; we'll just let nginx rewrite the path, and if the file doesn't exist, it should automatically generate a 404. That will be easier to maintain because to change it, all we need to do is start generating the missing file.

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -48,15 +54,44 @@ server {
...
+  }
+
+  location = /v4 {
+    types {
+      "application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4" json;
+    }
+    
+    if ($dcos_release_version != "1.10") {
+      rewrite ^/v4$ /repo-empty-v4.json break;
+    }
```
cruhland: Let's get rid of this special case too. Sorry!

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -48,15 +54,44 @@ server {
...
+
+    if ($universe_version = v4) {
+      rewrite ^/repo$ /v4;
     }
 
     rewrite ^/repo$ /repo-up-to-$dcos_release_version.zip break;
   }
 
+  location = /v3 {
```
cruhland: This looks like its defining an actual endpoint. Will clients be able to access `https://universe.marathon.mesos/v3`? Is there a way to prevent nginx from routing direct requests to this location?

scripts/gen-universe.py
```diff
@@ -59,18 +59,32 @@ def main():
...
+    """ Renders an empty universe json of with version `universe_version`
+
+    :param outdir: Path to the directory to use to store all universe objects
+    :type outdir: str
+    :param universe_version: Universe content type version: "v3" or "v4"
+    :type universe_version: str
+    :rtype: None
+    """
+    base_name = 'repo-empty-{}'.format(universe_version)
+    empty_path = outdir / (base_name + ".json")
```
cruhland: Why not include the `.json` in the string on the previous line? It should be easier to read.

cruhland: Oh never mind. I see, you reuse `base_name` with a different extension.

cruhland:
@larioj almost ready, just a few comments left.

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
larioj: Yeah, I think we can make this private.

scripts/gen-universe.py
larioj: we use base_name elsewhere. We also append '.content_type' to base_name

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
larioj: We just add the `internal` keyword.

PR #1041, opened 2017-03-24
Remove LD_LIBRARY_PATH variable and reduce the minVersion back to 1.8.

ssk2:
This solves a bug where the executor's LD_LIBRARY_PATH on DC/OS 1.8.8 would be improperly overwritten by the variable we had set in the marathon.json for Jenkins. This would cause Jenkins to fail to start on operating systems where libssl was not present in the system's PATH (e.g. CentOS but not CoreOS). 

This variable is no longer necessary - it is a legacy configuration option that we exposed to allow users to run this Jenkins package on non DC/OS Mesos clusters. Given that the Universe does not support non DC/OS clusters anymore, this is not a use case we need to support anymore.

Merge after: https://github.com/mesosphere/dcos-jenkins-service/pull/138 is merged and a new version of the Jenkins Docker image is released.

### DIFF
--- a/repo/packages/J/jenkins/17/resource.json
+++ b/repo/packages/J/jenkins/17/resource.json
@@ -7,7 +7,7 @@
   "assets": {
     "container": {
       "docker": {
-        "jenkins-302-2322": "mesosphere/jenkins:3.0.2-2.32.2"
+        "jenkins-302-2322": "mesosphere/jenkins:3.0.3-2.32.3"
       }
     }
   }
--- a/repo/packages/J/jenkins/17/resource.json
+++ b/repo/packages/J/jenkins/17/resource.json
@@ -7,7 +7,7 @@
   "assets": {
     "container": {
       "docker": {
-        "jenkins-302-2322": "mesosphere/jenkins:3.0.3-2.32.3"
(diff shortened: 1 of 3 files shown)


colin-msphere:
Has this change been tested against 1.8 and 1.9?

ssk2:
@colin-msphere: tested on 1.8 CentOS, will test on 1.8 CoreOS and 1.9 CoreOS.

repo/packages/J/jenkins/17/resource.json
```diff
@@ -7,7 +7,7 @@
   "assets": {
     "container": {
       "docker": {
-        "jenkins-302-2322": "mesosphere/jenkins:3.0.2-2.32.2"
+        "jenkins-302-2322": "mesosphere/jenkins:3.0.3-2.32.3"
```
colin-msphere: This should be `"jenkins-303-2323"` to match the `image` resource in the Marathon template.

ssk2: Argh, sorry. I made the change but didn't hit save.

ssk2: Updated now.

PR #1057, opened 2017-04-04
Several deployment improvements for docker local-universe 

mnp:
Two changes to `local-universe.py`, including README sections for each:

1. Work around the FAQ issue with local CLI and image assets being unavailable outside the cluster.  I added `--nonlocal_cli` and `--nonlocal_images` switches.  They will pass through whatever URL is specified in the resource.json without mapping it to an internal cluster URL. They will also inhibit fetching and caching of the resource.  This lets you host those assets elsewhere if needed.

2. Add a `--server_url` switch. This allows running a universe at an arbitrary internal URL, as a Marathon service for example, not just on masters.  The resource mapping stores this server URL as the host.  For example, I use http://dev-universe.marathon.mesos:8085 for my local universe.


mnp:
@jsancio thanks.  In our use case, we didn't need that, but I could see someone else needing it.  I'd be happy to add that to this PR for consistency if you think there's value.

PR #1083, opened 2017-04-12
Fix Kafka scheduler port designation in Marathon

nickbp:
New releases of the `confluent-kafka` and `kafka` packages which just update the packaging. The Marathon 'port' values should always be `0` to get a random port. If it's set to `1`, Marathon will want to reserve port 1.

```
packages/C/confluent-kafka$ diff -r 16/ 15/
diff -r 16/marathon.json.mustache 15/marathon.json.mustache
202c202
<       "port": 0,
---
>       "port": 1,
diff -r 16/package.json 15/package.json
4c4
<   "version": "1.1.19.1-3.2.0e",
---
>   "version": "1.1.19-3.2.0e",
```

```
packages/K/kafka$ diff -r 39/ 38/
diff -r 39/marathon.json.mustache 38/marathon.json.mustache
195c195
<       "port": 0,
---
>       "port": 1,
diff -r 39/package.json 38/package.json
16c16
<   "version": "1.1.19.1-0.10.1.0"
---
>   "version": "1.1.19-0.10.1.0"
```


nickbp:
cc @dbtucker

dbtucker:
LGTM ... thanks for catching this

dbtucker:
QQ on current framework: I didn't see the "health" port used at all ... only the "api" port (which was incorrectly initialized per the initial bug).  So I suspect that this would have kept multiple instances of Kafka tasks from being deployed on the same DC/OS host.
     Is that correct ?
     If so, does the change imply that single DC/OS hosts can support multiple Kafka tasks ?

nickbp:
This port just affects the scheduler, FWIW, so it should be unrelated to broker deployment. More recent versions of the service drop this second port.

PR #1111, opened 2017-04-20
marathon-lb: Bump to 1.7.0

nlsun:


nlsun:
@ryadav88 ready for review

nlsun:
@kensipe ready for review

repo/packages/M/marathon-lb/19/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "packagingVersion": "2.0",
```
kensipe: I'm not sure why this is still version 2.0.  this is very old.  version 3.0 is the latest and we are moving to 4 soon.

repo/packages/M/marathon-lb/19/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "marathon-lb",
+  "version": "1.7.0",
+  "scm": "https://github.com/mesosphere/marathon-lb",
+  "description": "HAProxy configured using Marathon state",
+  "maintainer": "support@mesosphere.io",
+  "tags": ["loadbalancer", "service-discovery", "reverse-proxy", "proxy", "haproxy"],
+  "preInstallNotes": "We recommend at least 2 CPUs and 1GiB of RAM for each Marathon-LB instance. \n\n*NOTE*: ```Enterprise Edition``` DC/OS requires setting up the Service Account in all security modes. \nFollow these instructions to setup a Service Account: https://docs.mesosphere.com/administration/id-and-access-mgt/service-auth/mlb-auth/",
```
kensipe: I understand you are making a simple version update.   Making a minor non-blocking comment here.   It probably makes sense to mention in the pre install notes that the 2 CPU, 1G recommendation is on the public node and that a public node must be available by default. (but perhaps no one reads these pre-install notes:)

nlsun:
@kensipe Actually there is a diff, I separated it out into 2 commits.

repo/packages/M/marathon-lb/19/package.json
nlsun: That's true, we can address this in a separate patch

repo/packages/M/marathon-lb/19/package.json
nlsun: That's true, we can address that in a separate patch

nlsun:
@ryadav88 We can address those in a separate patch.

PR #1147, opened 2017-05-03
Fix Accept header matching logic

jsancio:


jsancio:
@larioj @cruhland Manually tested both local universe and metadata only docker image.

PR #1148, opened 2017-05-04
extends jenkins configuration to be able to set jenkins port and jnlp…

zutherb:
we need that the Jenkins port and the JNLP port is configurable with the options. I added this


ssk2:
@zutherb: Thank you for the contribution! Can you walk me through why this is necessary in your use case? We usually try to avoid statically configuring ports because of how tasks are scheduled on Mesos (they can be anywhere).

zutherb:
hi @ssk2, we needed the jnlp port configurable because it seem there is a problem with the port auto discovery when running multi jenkins master on the same mesos slave. In this case both jenkins master got the port 50000 for the jenkins slave communication. 

The Jenkins port must be configurable because we want to use Amazon ALB to expose the Jenkins Masters like you got do it with a Kubernetes ingress. I miss a similar feature in mesos. 

Cheers 
Bernd

ssk2:
@zutherb: I assume you're using DC/OS for this? There should not be any problem with running multiple Jenkins masters on the same Mesos agent. It ordinarily randomly assigns a JNLP port for each outbound connection for a Jenkins agent.

Before we take this PR further, would you mind sharing your Jenkins configuration with us? I just want to double check your settings to see if there's a way we can resolve this without changing the package.

Ordinarily we expose Jenkins to the public using Marathon-lb on a DC/OS cluster. This will allow you to expose multiple Jenkins instances behind a single load balancer.

zutherb:
Hi @ssk2,

but please let's discuss this two things separately.

1. JNLP port:

This is only a workaround. Try to get two Jenkins Master on the same machine. We are not able to connect the Master with JNLP because both containers open port 50000.  Because JENKINS_SLAVE_AGENT_PORT is always set to 50000 in the original Jenkins container (https://github.com/jenkinsci/docker/blob/master/Dockerfile#L6 , https://github.com/jenkinsci/docker/blob/master/init.groovy#L9). 

2. Marathon LB

Thx for the recommendation. I agree with you In this case and we should change this. I thought the marathon-lb is deprecated but only haproxy-marathon-bridge. https://mesosphere.github.io/marathon/docs/service-discovery-load-balancing.html

repo/packages/J/jenkins/18/config.json
```diff
@@ -0,0 +1,142 @@
...
+                "virtual-host": {
+                    "description": "The virtual host address to configure for integration with Marathon-lb.",
+                    "type": "string"
+                },
+                "https-redirect": {
+                    "description": "Whether Marathon-lb should redirect HTTP traffic to HTTPS. This requires 'virtual-host' to be set. By default, this is false.",
+                    "type": "boolean",
+                    "default": false
+                },
+                "jenkins-slave-agent-port": {
```
colin-msphere: Please rename to `jenkins-agent-port`.

repo/packages/J/jenkins/18/config.json
```diff
@@ -0,0 +1,142 @@
...
+                    "description": "Optional argument to pass jenkins slave agent port to Jenkins",
```
colin-msphere: How about:

> Optional argument to set the agent port.

repo/packages/J/jenkins/18/config.json
```diff
@@ -0,0 +1,142 @@
...
+                    "type": "number"
```
colin-msphere: This should be `integer`.

zutherb:
@colin-msphere i did your changes

zutherb:
@colin-msphere i see that you requested changes but i can see what i should change?

colin-msphere:
@zutherb - are you interested in picking up this PR again? It will need a rebase and target version `22` instead of `18`. Otherwise, this can be closed and I will get these changes wrapped into the next Jenkins Universe package.

zutherb:
Hi @colin-msphere it did all of this a half year ago. The world has changed in the meanwhile and i am a little bit frustrated because of this PR. I guess it is better to close the PR. I don't have the time to check if this is needed anymore or if the bug was closed and to do the work.

PR #1183, opened 2017-05-23
Nifi - bump to v0.1, correct to Marathon-LB only, correct a few params

fernandosanchezmunoz:
- Nifi does not work properly over Admin Router. Removed the AR endpoint and added a fixed MLB endpoint
- Added a proper portDefinition to Marathon JSON that was missing 
- bumped package version to 0.1 from 0.0.1
- add configurable service port for MLB
- remove configurable "web-port" that was not being used
- cosmetic changes in doc

cc @tkrausjr @us3r

### DIFF
--- a/repo/packages/N/nifi/1/config.json
+++ b/repo/packages/N/nifi/1/config.json
@@ -1,0 +1,84 @@
+{
+    "type": "object",
+    "properties": {
+        "service": {
+            "description": "Configuration properties for the NIFI service for DC/OS.",
+            "type": "object",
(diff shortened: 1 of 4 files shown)


ryadav88:
@fernandosanchezmunoz please update the `dcos/example` as it still says you can configure it via AR. Apart from that this PR LGTM! Tested with mlb.

fernandosanchezmunoz:
@ryadav88 corrected here: https://github.com/dcos/examples/pull/174

PR #1185, opened 2017-05-24
dcos-enterprise-cli: update minDcosReleaseVersion to 1.9

tamarrow:


repo/packages/D/dcos-enterprise-cli/12/package.json
```diff
@@ -3,7 +3,7 @@
   "name": "dcos-enterprise-cli",
   "version": "1.0.8",
   "maintainer": "support@mesosphere.io",
-  "minDcosReleaseVersion": "1.8",
+  "minDcosReleaseVersion": "1.9",
```
vespian: Aren't we worried that we are changing existing/already released package here? Also - why this change in general?

tamarrow: we have to update because if we add a new package that restricts to 1.9 only 1.8 clusters will still get package 12, which we don't want. 

the endpoint here only exists in 1.9+ (which is the reason for this change):  https://github.com/mesosphere/dcos-security-cli/blob/master/dscli/__init__.py#L62

PR #1188, opened 2017-05-26
Updated redis version to 3.2.9.

joerg84:


joerg84:
Just got verbal approve from ravi and I just retested it with the DC/OS 101 tutorial...

PR #1203, opened 2017-06-01
Adding an enum for validation

triclambert:


repo/packages/B/beta-kafka/2/config.json
```diff
@@ -43,6 +43,7 @@
                 "beta-optin": {
                     "title": "Agree to Beta terms",
                     "type": "boolean",
+                    "enum": ["true"],
```
jsancio: I think you want `"enum": [true]`. `"true"` is a string not a boolean.

PR #1217, opened 2017-06-08
fixed hdfs service name

ANierbeck:
The used hdfs-site.xml does point to the incorrect URL for the hdfs-servers, that's because the template isn't filtered correctly. 
This PR should fix that.


nickbp:
This was fixed upstream via https://github.com/mesosphere/dcos-commons/pull/1066

Additional checks are being added in this pending upstream PR, where the service will throw on startup if template parameters are missing, instead of proceeding silently: https://github.com/mesosphere/dcos-commons/pull/1121/files#diff-138d0da5741f190fa5bfcf1254092f72
The PR also adds some integration tests which check the resulting output: https://github.com/mesosphere/dcos-commons/pull/1121/files#diff-6f04999201ffbed4da39b4d5cc50cd19R47

The next HDFS cut this week will have the above fixes.

PR #1218, opened 2017-06-08
Fix spark history

ANierbeck:
Fixes the spark history server to work in a Offline Repository environment. 
Especially since a build of a Offline Repository wasn't possible. 
It's also now possible to install the spark history server from the dcos ui, previously it's only been possible to do that via CLI

### DIFF
--- a/repo/packages/S/spark-history/0/config.json
+++ b/repo/packages/S/spark-history/0/config.json
@@ -1,62 +1,73 @@
 {
-    "type": "object",
-    "properties": {
+  "type": "object",
+  "properties": {
+    "service": {
+      "type": "object",
+      "description": "DC/OS service configuration properties",
+      "properties": {
         "name": {
-            "default": "spark-history",
-            "description": "The app name for the Spark History Server.  The service will be available at http://<dcos_url>/service/<name>/",
-            "type": "string"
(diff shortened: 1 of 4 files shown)


mgummelt:
Thanks for the PR, but as with all data services packages, the source is elsewhere.  This is just object code.  You'll need to retarget your PR to https://github.com/mesosphere/spark-build/tree/master/history/package

ANierbeck:
@mgummelt, thanks for the info. Is there some documentation about this information, couldn't find it? 
Another question, will this PR changes somehow merged into the spark-build project?

mgummelt:
This PR will not be auto-merged into spark-build.  To ultimately get these changes into universe, you need to resubmit your PR to the spark-build project, after which I will push the updated package to universe.

mgummelt:
There's no explicit mapping for where the source exists for each package, no.  That would be a nice improvement to universe, though.

ANierbeck:
@mgummelt ... where do I need to file such a feature request :)

mgummelt:
http://jira.dcos.io 

Thanks!

ANierbeck:
@mgummelt see https://jira.mesosphere.com/browse/DCOS_OSS-1306

PR #1219, opened 2017-06-08
added ceph url in resource.json

ANierbeck:
This should also make it possible to work with an offline repository containing the cephs packages.

### DIFF
--- a/repo/packages/C/ceph/2/resource.json
+++ b/repo/packages/C/ceph/2/resource.json
@@ -5,6 +5,9 @@
     "icon-large": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-large.png"
   },
   "assets": {
+    "uris": {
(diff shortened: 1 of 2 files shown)


OhRobin:
@ryadav88 this looks good to me

repo/packages/C/ceph/2/resource.json
```diff
@@ -5,6 +5,9 @@
     "icon-large": "https://s3.amazonaws.com/downloads.mesosphere.io/universe/assets/icon-service-ceph-large.png"
   },
   "assets": {
+    "uris": {
+      "ceph-tgz": "https://dl.bintray.com/vivint-smarthome/ceph-on-mesos/ceph-on-mesos-0.2.11.tgz"
```
jsancio: @ANierbeck We have noticed that this URL is return a `403` for GET requests.

jsancio: cc @takirala

PR #1221, opened 2017-06-09
Spark thrift: remove redundant parameter and add configurable spark image

fernandosanchezmunoz:
there is a repeated parameter in config.json "spark-max-executors", this is just the deletion of the redundancy.


ryadav88:
@jsancio @cruhland can you check why these builds are failing?

jsancio:
Re-ran build. It should be fixed now.

PR #1226, opened 2017-06-13
Add hdfs and update helloworld packages

jsancio:
We are making this changes so we can:

1. Get rid of the helloworld.zip used by Cosmos' integration tests. To
get rid of that universe repo we need to add the packages hdfs and
helloworld/0.

2. Some of our integration tests that use helloworld override the
Marathon app id for the package by using the optional app id in the
install request. We shouldn't do this for two reasons:
  a. That is a deprecated feature that we want to stop supporting.
  b. It makes it impossible to update by using the new service update
  RPC.
For the helloworld packages we are introducing a new "name" property
that we can use to override the app id for the package.


jsancio:
Thanks for the review @cruhland !

PR #1227, opened 2017-06-13
Document and streamline update process for universe-static image

cruhland:
Trying to provide some help to whoever will need to update this in the future.

### DIFF
--- a/docker/local-universe/Makefile
+++ b/docker/local-universe/Makefile
@@ -1,5 +1,5 @@
 REPO_BASE_DIR ?= $(shell pwd -P)/../..
-static_version ?= 0.2
+static_version ?= 0.2-2
 static_image ?= universe-static:$(static_version)
 
 .PHONY: certs base clean
--- a/docker/local-universe/Makefile
+++ b/docker/local-universe/Makefile
@@ -1,4 +1,7 @@
 REPO_BASE_DIR ?= $(shell pwd -P)/../..
+static_version ?= 0.2
+static_image ?= universe-static:$(static_version)
+
 .PHONY: certs base clean
 
 certs:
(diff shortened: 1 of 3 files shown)


docker/local-universe/Makefile
```diff
@@ -1,4 +1,7 @@
 REPO_BASE_DIR ?= $(shell pwd -P)/../..
+static_version ?= 0.2
```
cruhland: This can be overridden on the command-line by doing e.g. `make static_version=0.3 <target>`

docker/local-universe/README.md
```diff
@@ -84,7 +84,7 @@ Mesosphere provides a `mesosphere/universe-static` Docker image, which has all o
 
     ```bash
     $ sudo make static-online
-    ## Will pull and re-tag mesosphere/universe-static:0.2 to universe-static
```
cruhland: Explicit version isn't needed in an explanatory comment

docker/local-universe/README.md
```diff
@@ -190,3 +190,13 @@ For example, if you want dev-universe, then add `--server_url http://dev-univers
...
 ```
+
+## Making changes...
+
+###  ...to the `universe-static` image
+
+1. Update `Dockerfile.static` with the changes
+1. Bump the default value of `static_version` in `Makefile`
+1. Create a pull request with the above changes
+1. Once it is merged, run the [Publish universe-static image](https://teamcity.mesosphere.io/viewType.html?buildTypeId=Oss_Universe_2_PublishUniverseStaticImage&branch_Oss_Universe_2=%3Cdefault%3E&tab=buildTypeStatusDiv) build, providing the new version number when prompted
```
cruhland: I haven't updated the build steps yet, because one of the steps actually gets more verbose with this change! It's because the `Makefile` doesn't build the image with the `mesosphere/` prefix, so we have to tag it separately. Here's the current build step:

```bash
#!/bin/bash
set -o errexit -o nounset -o pipefail

docker build -t "mesosphere/universe-static:%docker.image.version%" -f Dockerfile.static .

exit
```

And here's what we'd have to update it to:
```bash
#!/bin/bash
set -o errexit -o nounset -o pipefail

make static_version=%docker.image.version% static-build
docker tag universe-static:%docker.image.version% "mesosphere/universe-static:%docker.image.version%"

exit
```

@jsancio any suggestions?

jsancio: I see. I would go a step further and use the same pattern we use in Cosmos. What do you think about `universe-static:<version>-<build-number>`?

cruhland: Oh, I thought we were only going to run the build once per version. There wouldn't be any need for a build number in that case, right? Or do you want to have this run on every commit to the `version-3.x` branch?

jsancio: I am concerned with human error. We are not stopping anyone from running that CI twice, right?

jsancio:
LGTM!

PR #1232, opened 2017-06-16
Allow local universe to be built for a specific DC/OS version

cruhland:
To do so, pass the command-line argument `--dcos_version VERSION` to the `local-universe.py` script. For example, `--dcos_version 1.9` would ensure that for each included package, only the most recent revision that could run on DC/OS 1.9 would be what ends up in the local universe image. See the documentation in the diff below for more information.

Fixes https://jira.mesosphere.com/browse/DCOS-16231.

### DIFF
--- a/scripts/local-universe.py
+++ b/scripts/local-universe.py
@@ -3,6 +3,7 @@
 import argparse
 import concurrent.futures
 import contextlib
+import distutils.version
 import fnmatch
 import json
 import os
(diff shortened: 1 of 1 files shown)


scripts/local-universe.py
```diff
@@ -62,11 +63,21 @@ def main():
...
+        'that DC/OS version are included. If unset, the latest version of '
+        'each package will be included.'
+    )
 
     args = parser.parse_args()
 
     package_names = [name for name in args.include.split(',') if name != '']
 
+    dcos_version = distutils.version.LooseVersion(args.dcos_version) \
+        if args.dcos_version is not None else None
```
jsancio: Okay but if you want `... if args.dcos_version else None`. In Python `None` and `''` (empty string) are `False` in an `if` statement.

cruhland: Oh good point, an empty string would be just as bad as `None`. I'll update.

PR #1237, opened 2017-06-21
[INFINITY-1785] Remove beta-optin option

triclambert:
The beta-optin is proving confusing for end-users.  Let's remove it in favor of a different approach.

### DIFF
--- a/repo/packages/B/beta-cassandra/2/config.json
+++ b/repo/packages/B/beta-cassandra/2/config.json
@@ -53,18 +53,11 @@
                         "parallel"
                     ],
                     "default": "serial"
-                },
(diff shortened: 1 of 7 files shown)


benclarkwood:
@triclambert Can you please create a PR (or a JIRA) for removing the automatic insertion of this in the release process? Otherwise this change will immediately be undone.

benclarkwood:
@triclambert Why not beta-confluent?

Also, we are about to release new beta versions and they will not have the beta-optin in them.

triclambert:
Whoops, forgot Confluent.  Added.  

Looks like you already updated release_builder.py.  Thanks.

PR #1254, opened 2017-06-28
Add Chronos 2.5.0.

gkleiman:


ryadav88:
@gkleiman Can we review https://github.com/dcos/examples/tree/master/chronos/1.9 and see if this needs edits based on your latest PR?

gkleiman:
If we merge this PR, then users will be able to choose between Chronos v2.5.0 and v3.0.1.

The tutorial is for Chronos v3.0.1, so it doesn't need to change unless we decide to recommend users to use Chronos v2.50 instead of v3.0.1.

repo/packages/C/chronos/5/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "packagingVersion": "4.0",
+  "name": "chronos",
+  "version": "2.5.0",
+  "scm": "https://github.com/mesos/chronos.git",
+  "maintainer": "support@mesosphere.io",
+  "description": "A fault tolerant job scheduler for Mesos which handles dependencies and ISO8601 based schedules.",
+  "framework": true,
+  "minDcosReleaseVersion": "1.9",
```
ryadav88: @gkleiman you only want this to be available in `minDcosReleaseVersion ` >= 1.9 ?

gkleiman: Yes, we tested Chronos on the 1.9 on a soak cluster, but not on 1.8.

PR #1258, opened 2017-06-29
marathon-lb: Bump to v1.8.0

nlsun:
https://jira.mesosphere.com/browse/DCOS_OSS-1312


nlsun:
@ryadav88 please review. Seems like @kensipe was tagged before as well. Who should I be asking for reviews?

PR #1265, opened 2017-07-07
Make --dcos_version parameter required for building local Universe

cruhland:
Addresses [DCOS_OSS-1299](https://jira.mesosphere.com/browse/DCOS_OSS-1299).

This will ensure that all packages in a local Universe image are compatible with the cluster they will be used in.

I'll have a PR to update the DC/OS documentation shortly. We might want to wait to merge this until after the docs are updated.

### DIFF
--- a/docker/local-universe/Makefile
+++ b/docker/local-universe/Makefile
@@ -5,16 +5,16 @@ static_image ?= universe-static:$(static_version)
 .PHONY: certs base clean
 
 certs:
-	mkdir certs && openssl req					\
-		-newkey rsa:4096 -nodes -sha256 -keyout certs/domain.key	\
-		-x509 -days 365 -out certs/domain.crt			\
+	mkdir certs && openssl req \
+		-newkey rsa:4096 -nodes -sha256 -keyout certs/domain.key \
+		-x509 -days 365 -out certs/domain.crt \
 		-subj "/CN=master.mesos"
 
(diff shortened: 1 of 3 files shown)


justinrlee:
Just a heads up - some users like to use a single local-universe for multiple DC/OS clusters (which may be on different DC/OS versions).  Is the --dcos-version a minimum version or max version?

cruhland:
@justinrlee the parameter to the script is an exact version match, but the packages themselves specify the minimum DC/OS version they are compatible with. As an example, let's say we have package A with min version 1.8, package B with min version 1.9, and package C with min version 1.10. Here's what would be included in the local universe image for the following parameter values:
* `--dcos_version 1.7`: nothing (all packages have min versions > 1.7)
* `--dcos_version 1.8`: package A
* `--dcos_version 1.9`: packages A and B
* `--dcos_version 1.10`: packages A, B, and C

It's actually a bit more complicated, because each package revision can specify a different min DC/OS version. For that reason, making the `--dcos_version` parameter required is very important -- the customers using a single local Universe image on clusters of different versions will observe missing packages on some clusters, if the versions are different enough. Unfortunately we don't have a way to add multiple revisions of a package to a single image, but @jsancio is scheduled to implement that in the current sprint (see [DCOS-16635](https://jira.mesosphere.com/browse/DCOS-16635)).

PR #1271, opened 2017-07-17
enhances validations

MrMarvin:
With this commit we will be able to run a certain set of validation
only against the most recent revision of a package. This is helpful to
not dig in historic, non-changeable files for (now) unsupported features.

Two checks are provided with this feature:
Check for non-https links and
validate that assets hosted on downloads.mesosphere.com are in a specific sub-directory.
The later only applies to packages hosted by mesosphere and we should use these WARNINGS
to lean up our distribution.
Suggestion: Go trough package by package and fix what is needed in a new revision each.

I am open for discussion on if and how we should proceed with this. 
Reference: https://jira.mesosphere.com/browse/QUALITY-1487

### DIFF
--- a/scripts/validate-packages.py
+++ b/scripts/validate-packages.py
@@ -5,6 +5,7 @@ import jsonschema
 import os
 import sys
 from distutils.version import LooseVersion
+from urllib.parse import urlparse
 
 SCRIPTS_DIR = os.path.dirname(os.path.realpath(__file__))
 UNIVERSE_DIR = os.path.join(SCRIPTS_DIR, "..")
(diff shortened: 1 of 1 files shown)


MrMarvin:
After looking at the Teamcity builds I realized that there is a separate repo with 'A suite of tests and tools that run agains PRs to https://github.com/mesosphere/universe'. So I guess the checks in this universe repo itself are more generic non-mesosphere specific and I should move my Mesosphere specific sanity checks to the other repo and/or change the teamcity build?

scripts/validate-packages.py
```diff
@@ -133,6 +138,22 @@ def _validate_revision(given_package, revision, path):
...
+            all_good = False
+            eprint("\n\t\t\tWARNING non-https link: {}".format(uri), end='')
+
+        # check for links on downloads downloads.mesosphere.com that do not point
+        # to a dedicated sub dir.
+        if (urlparse(uri).hostname == 'downloads.mesosphere.com' and not urlparse(uri).path.startswith('/universe/assets/') ):
+            all_good = False
+            eprint("\n\t\t\tWARNING downloads.mesosphere.com asset in stray directory: {}".format(uri), end='')
+
+    return all_good
```
jsancio: Regarding the two warnings above, we should turn them to errors and `sys.exit`. Developers are not going to notice those warning and it wont have the effect you want.

MrMarvin: I was planing on using these warnings in the first step, then going through the packages myself and correct these for each package with a new revision. This is supposed to be a soft launch of a few rules:
* only https links
* consolidated path to packages for easier replication of a whole universe

After that we should make it hard errors. wdyt?

Also, whats your opinion on my comment above that this is very mesosphere specific and maybe should go in the other checker repository?

jsancio: @MrMarvin excuse the delay response. I must have missed this.

Yeah. Let's move this to the other repo.

I understand now what you are trying to do. We can leave it as warnings for now. Before turning them to error we need to communicate this to the rest of the organization.

ryadav88:
@MrMarvin @jsancio can this PR be closed?

MrMarvin:
> can this PR be closed?

Yep, sorry for the delay! I'll see to get the chance to move these checks over to our mesosphere specific checks as discussed. Closing this!

PR #1275, opened 2017-07-18
Updated package schema for stringent validation of service name

takirala:
Current rules for validation are : 

- Starts with [a-z]
- each character can only be [a-z] or - or [0-9]

### DIFF
--- a/repo/meta/schema/package-schema.json
+++ b/repo/meta/schema/package-schema.json
@@ -27,7 +27,7 @@
         },
         "name": {
           "type": "string",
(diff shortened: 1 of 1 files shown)


repo/meta/schema/package-schema.json
```diff
@@ -27,6 +27,7 @@
         },
         "name": {
           "type": "string",
+          "pattern": "^[a-z]+[a-z0-9-]*[a-z0-9]+$",
```
cruhland: Let's make sure these match your validation script: `^[a-z][a-z0-9-]*[a-z0-9]$`

PR #1277, opened 2017-07-18
Local package versions

jsancio:
Tested this manually by running the following commands:

```bash
sudo make DCOS_VERSION=1.10 local-universe
```

```bash
sudo make DCOS_VERSION=1.10 DCOS_PACKAGE_INCLUDE="cassandra:1.0.25-3.0.10,cassandra:1.0.24-3.0.10,marathon:1.4.2" local-universe 
```

### DIFF
--- a/docker/local-universe/README.md
+++ b/docker/local-universe/README.md
@@ -103,7 +103,7 @@ something more than that.
     $ sudo make DCOS_VERSION=<your DC/OS version> DCOS_PACKAGE_INCLUDE="cassandra:1.0.25-3.0.10,cassandra:1.0.24-3.0.10,marathon:1.4.2" local-universe
     ```
 
-## Building Your Own, off a non-changing universe-static base image.
+### Building Your Own, off a non-changing universe-static base image.
 
 Mesosphere provides a `mesosphere/universe-static` Docker image, which has all of the core
 requirements to run a local universe.  All that must be added are your own certificates and repo
@@ -150,6 +150,16 @@ make DCOS_VERSION=<your DC/OS version> local-universe
 make static-build
 make static-base
 make DCOS_VERSION=<your DC/OS version> local-universe
+```
+
+### Building an old version of Local Universe
+The latest version of Local Universe has changed directory structure in a way that is not
+compatible with the old version of Local Universe. If you are interested in creating an old version
+of Local Universe with the old directory structure you must use the following command instead of
+the `local-universe` make target describe previously.
+
+```bash
+$ sudo make DCOS_VERSION=<your DC/OS version> old-local-universe
 ```
 
 ### Outside Resources
--- a/docker/local-universe/README.md
+++ b/docker/local-universe/README.md
@@ -1,7 +1,10 @@
 
 ## Using
 
-1. If you're using the latest stable release of DC/OS, then download the [local-universe](https://downloads.mesosphere.com/universe/public/local-universe.tar.gz) container to each of your masters. Otherwise, you'll need to build the container for your DC/OS version (see the "Building Your Own" section below) and copy it to each of your masters.
+1. If you're using the latest stable release of DC/OS, then download the [local-universe]
+(https://downloads.mesosphere.com/universe/public/local-universe.tar.gz) container to each of your
+masters. Otherwise, you'll need to build the container for your DC/OS version (see the "Building
+Your Own" section below) and copy it to each of your masters.
 
 1. Load the container into the local docker instance on each master:
 
@@ -9,7 +12,8 @@
     $ docker load < local-universe.tar.gz
     ```
 
-1. Add the [`dcos-local-universe-http.service`](dcos-local-universe-http.service) definition to each of your masters at `/etc/systemd/system/dcos-local-universe-http.service` and then start it.
+1. Add the [`dcos-local-universe-http.service`](dcos-local-universe-http.service) definition to
+each of your masters at `/etc/systemd/system/dcos-local-universe-http.service` and then start it.
 
     ```bash
     $ cp dcos-local-universe-http.service /etc/systemd/system/dcos-local-universe-http.service
@@ -17,7 +21,9 @@
     $ systemctl start dcos-local-universe-http
     ```
 
-1. Add the [`dcos-local-universe-registry.service`](dcos-local-universe-registry.service) definition to each of your masters at `/etc/systemd/system/dcos-local-universe-registry.service` and then start it.
+1. Add the [`dcos-local-universe-registry.service`](dcos-local-universe-registry.service)
+definition to each of your masters at `/etc/systemd/system/dcos-local-universe-registry.service`
+and then start it.
 
     ```bash
     $ cp dcos-local-universe-registry.service /etc/systemd/system/dcos-local-universe-registry.service
@@ -25,7 +31,8 @@
     $ systemctl start dcos-local-universe-registry
     ```
 
-1. Remove the built in repositories from the host that you have the DCOS-CLI installed on (alternatively, these can be removed from the DCOS UI under System>Repositories).
+1. Remove the built in repositories from the host that you have the DCOS-CLI installed on
+(alternatively, these can be removed from the DCOS UI under System>Repositories).
 
     ```bash
     $ dcos package repo remove Universe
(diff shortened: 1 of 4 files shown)


docker/local-universe/README.md
```diff
@@ -1,31 +1,38 @@
 
 ## Using
 
-1. If you're using the latest stable release of DC/OS, then download the [local-universe](https://downloads.mesosphere.com/universe/public/local-universe.tar.gz) container to each of your masters. Otherwise, you'll need to build the container for your DC/OS version (see the "Building Your Own" section below) and copy it to each of your masters.
+1. If you're using the latest stable release of DC/OS, then download the [local-universe]
+(https://downloads.mesosphere.com/universe/public/local-universe.tar.gz) container to each of your
+masters. Otherwise, you'll need to build the container for your DC/OS version (see the "Building
+Your Own" section below) and copy it to each of your masters.
```
cruhland: @jsancio it looks like most of these changes are formatting only. Could you add a comment to the places where the content actually changed?

scripts/local-universe.py
```diff
@@ -45,9 +45,9 @@ def main():
     parser.add_argument(
         '--include',
         default='',
-        help='Command separated list of packages to include. If this option '
-        'is not specified then all packages are downloaded. E.g. '
-        '--include="marathon,chronos"')
+        help='Command (,) separated list of packages to include. For each '
```
cruhland: Command -> Comma?

cruhland: Why did you drop the help text for when the option isn't specified?

scripts/local-universe.py
```diff
@@ -45,9 +45,9 @@ def main():
...
+        'package specify both the name and version by separating them with a '
+        'colon (:). E.g. --include="marathon:1.4.2,chronos:2.5.0"')
```
cruhland: Should we support optional versions, for backwards compatibility?

cruhland: Hmm, actually, we probably don't want that, since it wouldn't get users to switch to using versions.

scripts/local-universe.py
```diff
@@ -153,52 +158,27 @@ def handle_package(opts):
...
     """Enumarate all of the package and revision to include
 
     :param packages_path: the path to the root of the packages
     :type packages_path: pathlib.Path
-    :param package_names: list of package to include. empty list means all
-                         packages
-    :type package_names: [str]
+    :param packages: list of the name and version of packages to include. empty
+                     list means all packages
+    :type packages: [[str]]
```
cruhland: Convert the `[name, version]` list into a tuple, so we can use `[(str, str)]` here?

cruhland: I know Python's tuples are just immutable lists, but being able to have a more specific type is useful.

scripts/local-universe.py
```diff
@@ -208,43 +188,100 @@ def include_revision(package_name, revision_path):
 
             # Include only the first acceptable revision
```
cruhland: This comment is out of date.

scripts/local-universe.py
```diff
@@ -208,43 +188,100 @@ def include_revision(package_name, revision_path):
...
-        resource = json.load(json_file)
+
+def selected_check(package_json, packages, only_selected):
+    package_name = package_json['name']
+    package_version = package_json['version']
+
+    if only_selected:
+        return package_json.get('selected', False)
+
+    return [package_name, package_version] in packages
```
cruhland: Did you drop the `not packages` condition because it would include too many packages?

scripts/local-universe.py
cruhland: Ah, I think I answered my own question in another comment below. There would be too many packages to include if "not specified" meant "all versions of all packages".

scripts/local-universe.py
```diff
@@ -319,35 +356,36 @@ def add_http_resource(dir_path, url, base_path):
...
+    with dest_resource.open('w', encoding='utf-8') as dest_file:
+        resource = load_json(package_path / 'resource.json')
 
         # Change the root for images (ignore screenshots)
         if not skip_images and 'images' in resource:
             resource["images"] = {
                 n: urllib.parse.urljoin(
                     http_root, str(pathlib.PurePath(
-                        package, "images", pathlib.Path(uri).name)))
+                        package, version, "images", pathlib.Path(uri).name)))
```
cruhland: You're including the version in all these paths, but I don't see where you're updating the directories on the filesystem to include the version directory.

cruhland: Oh, it's in the `enumerate_http_resources()` function, correct?

scripts/local-universe.py
justinrlee: I like the idea of optional versions.  Are we trying to force users to use versions?

cruhland: @justinrlee it's generally a good idea to use explicit versions, because it makes the build deterministic. It might seem like a benefit for a customer to get package updates automatically when they rebuild their local universe image, but I'd bet they'd prefer to know when the version is changing and what they need to do to handle upgrades safely.

docker/local-universe/README.md
```diff
@@ -38,47 +45,69 @@
...
+something more than that.
 
     ```bash
     $ sudo make DCOS_VERSION=<your DC/OS version> local-universe
     ```
 
-## Building Your Own, off a non-changing universe-static base image.
+    If you want to specify the set of package names and versions that get included in the image you
+    can user the `DCOS_PACKAGE_INCLUDE variable. For example, to include two versions for Cassandra
+    and one version of Marathon you can execute the following command.
```
jsancio: Wording changed here.

docker/local-universe/README.md
```diff
@@ -38,47 +45,69 @@
...
-Mesosphere provides a `mesosphere/universe-static` Docker image, which has all of the core requirements to run a local universe.  All that must be added are your own certificates and repo contents.
+    ```bash
+    $ sudo make DCOS_VERSION=<your DC/OS version> DCOS_PACKAGE_INCLUDE="cassandra:1.0.25-3.0.10,cassandra:1.0.24-3.0.10,marathon:1.4.2" local-universe
+    ```
+
+### Building Your Own, off a non-changing universe-static base image.
+
+Mesosphere provides a `mesosphere/universe-static` Docker image, which has all of the core
+requirements to run a local universe.  All that must be added are your own certificates and repo
+contents.
```
jsancio: Wording changed here.

docker/local-universe/README.md
```diff
@@ -121,30 +152,46 @@ make static-base
...
+### Building an old version of Local Universe
+The latest version of Local Universe has changed directory structure in a way that is not
+compatible with the old version of Local Universe. If you are interested in creating an old version
+of Local Universe with the old directory structure you must use the following command instead of
+the `local-universe` make target describe previously.
+
+```bash
+$ sudo make DCOS_VERSION=<your DC/OS version> old-local-universe
+```
+
```
jsancio: Added this section.

scripts/local-universe.py
jsancio: Done.

jsancio: Yes. I think the previous behavior was a mistake. We should avoid suggesting that user should generate a universe that contains all of the packages.

scripts/local-universe.py
jsancio: Yes. Indirectly, users have told us that they want to be able to produce their Local Universe and in the process possibly add new packages. The best way to do that is to require that packages are enumerated explicitly.

scripts/local-universe.py
jsancio: Done.

scripts/local-universe.py
jsancio: Fixed. Moved the documentation to the `selected_check` function.

scripts/local-universe.py
jsancio: Yeah. I ended up changing the semantic of an empty `packages`. We were interpreting as download everthing. I don't think that is what the user wants.

scripts/local-universe.py
jsancio: Yeah. Unfortunately, we do that in two places. In general, the functional separation of this script is not that great. I think it is okay for now since we plan to deprecate this script after 1.11.

jsancio:
@cruhland Addressed your comments and fix a bug when using `--selected`.

scripts/local-universe.py
```diff
@@ -45,9 +45,9 @@ def main():
     parser.add_argument(
         '--include',
         default='',
-        help='Command separated list of packages to include. If this option '
-        'is not specified then all packages are downloaded. E.g. '
-        '--include="marathon,chronos"')
+        help='Comma (,) separated list of packages to include. For each '
```
larioj: 👍

cruhland:
@jsancio :ship: it!

PR #1281, opened 2017-07-21
Added config option for resource role in Nexus.

joerg84:


joerg84:
Moved to new version:

diff 3/config.json 4/config.json
23a24,28
>         },
>         "resource-role": {
>           "default": "*",
>           "description": "The accepted resource role. Change to \"slave_public\" role if it should run on an public agent.",
>           "type": "string"


diff 3/marathon.json.mustache 4/marathon.json.mustache
4a5,6
>   "acceptedResourceRoles": [ "{{service.resource-role}}" ],
>

repo/packages/N/nexus/4/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "packagingVersion": "3.0",
+  "name": "nexus",
+  "version": "3.2.1",
```
ryadav88: please update the version.

ryadav88:
@whyjustin PTAL ^

repo/packages/N/nexus/4/package.json
whyjustin: There is a 3.4.0 version available. My team is about to embark on a project to automate the updates for this kind of change but until then we could update in this PR or I can submit a separate PR. Not sure your preference as separate PRs require a new folder in the tree.

joerg84: @whyjustin I would prefer a separate PR as then I can start working with my demo. Would be great if you could keep the config parameter as it makes it easy to expose nexus via the public agents in demo environments.

whyjustin: 👍

whyjustin:
> @whyjustin PTAL ^

LGTM

PR #1309, opened 2017-07-28
marathon-lb: Make minDcosVersion 1.10

nlsun:


nlsun:
@cruhland Thats a good question, but I'm not sure

Issue #945, opened 2017-02-06
Use of `jsonschema` and `jq` in build scripts brings in extra dependencies for repo consumers


dannypurcell:
The README section on [Repository Consumption](https://github.com/mesosphere/universe#repository-consumption-1) says one should be able to do
```
scripts/build.sh
DOCKER_TAG="my-package" docker/server/build.bash
```
however, doing so on a fresh centos box seems to require the installation of the following dependencies
* `sudo yum install epel-release`
* `sudo yum install python34`
* `sudo yum install python34-devel`
* `curl -O https://bootstrap.pypa.io/get-pip.py`
* `sudo /usr/bin/python3.4 get-pip.py`
* `pip install jsonschema`
* `sudo yum install jq`

This significantly complicates the process to merely get a usable copy of the universe repo image.

Additionally, since these dependencies are not mentioned in the README, new users currently have to discover them by banging their head against the console and digging around on the internet. This is likely to arouse hate and discontent among enterprise edition users who need to run a local copy of the Universe for security reasons.

Can we fix up this process so it's a bit nicer to the first time user?

joerg84:
@dannypurcell thx for your feedback! looking into this

mastermindg:
**Why has this not been corrected yet?** This is a simple fix. Just detail the dependencies in the README. It's not like people can't install them - it's just a BIG pain that the dependencies are not listed in the README and users need to come to issues to find this.

jsancio:
@dannypurcell and @mastermindg, excuse the delayed response. We are tracking this and should be scheduled to get worked on soon.

PR #1322, opened 2017-07-31
Update Readme and shell script to check for dependencies

takirala:
Make sure `jsonschema` is installed while validating repo packages. Also check for `jq` when building/publishing docker image. 

Addresses https://github.com/mesosphere/universe/issues/945


docker/server/build.bash
```diff
@@ -75,6 +75,12 @@ function gzipJsonFiles {(
 
 function build {(
 
+  # Check if the jq processor is installed correctly
+  if ! which jq > /dev/null 2>&1; then
```
cruhland: Use `! command -v jq` instead -- see [this SO answer](https://stackoverflow.com/a/677212/7002313).

scripts/build.sh
```diff
@@ -15,6 +15,11 @@ python3 -m venv ${REPO_BASE_DIR}/target/venv
 # Install dependencies
 ${REPO_BASE_DIR}/target/venv/bin/pip install -r ${SCRIPTS_DIR}/requirements/requirements.txt
 
+# Check if the dependencies were installed successfully
+if ! which jsonschema > /dev/null 2>&1; then
```
cruhland: Use `! command -v jsonschema` instead -- see [this SO answer](https://stackoverflow.com/a/677212/7002313).

scripts/build.sh
```diff
@@ -15,6 +15,11 @@ python3 -m venv ${REPO_BASE_DIR}/target/venv
 # Install dependencies
 ${REPO_BASE_DIR}/target/venv/bin/pip install -r ${SCRIPTS_DIR}/requirements/requirements.txt
 
+# Check if the dependencies were installed successfully
+if ! command -v jsonschema > /dev/null 2>&1; then
+  echo "jsonschema is required. Please install!"
+  exit 1
+fi
```
jsancio: This is not required. We install `jsonschema` automatically here: https://github.com/mesosphere/universe/blob/version-3.x/scripts/requirements/requirements.txt#L1 which is executed in the line 16 above. I'll create a PR.

PR #1330, opened 2017-08-03
Change the healthcheck protocol to HTTPS in case TLS is enabled

NitishT:


repo/packages/M/minio/8/config.json
```diff
@@ -0,0 +1,101 @@
...
+        "service": {
+            "type": "object",
+            "description": "Service properties for Minio on DC/OS",
+            "properties": {
+                "name": {
+                    "description": "Name of this service instance",
+                    "type": "string",
+                    "default": "minio"
+                },
+                "ssl": {
```
harshavardhana: ssl is a deprecated protocol i think we should call it TLS?

NitishT: yes, agreed. Fixed

NitishT:
@ryadav88 any reviews here?

harshavardhana:
Thanks @ryadav88

PR #1332, opened 2017-08-03
Add metadata and build definition schemas

larioj:
These are used by the dcos package build cli. It would be useful to keep them here, so that we update them whenever we update the other schemas.

### DIFF
--- a/repo/meta/schema/build-definition-schema.json
+++ b/repo/meta/schema/build-definition-schema.json
@@ -91,64 +91,6 @@
       "additionalProperties": false
     },
 
-
-    "v20resource": {
-      "additionalProperties": false,
-      "type": "object",
-      "properties": {
-        "assets": {
-          "type": "object",
-          "properties": {
-            "uris": {
-              "type": "object",
-              "additionalProperties": {
-                "type": "string"
-              }
-            },
-            "container": {
-              "type": "object",
-              "properties": {
-                "docker": {
-                  "type": "object",
-                  "additionalProperties": {
-                    "type": "string"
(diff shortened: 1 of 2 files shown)


repo/meta/schema/build-definition-schema.json
```diff
@@ -0,0 +1,611 @@
...
+      "additionalProperties": false
+    }
+
+  },
+
+  "type": "object",
+  "oneOf": [
+    { "$ref": "#/definitions/v20Package" },
+    { "$ref": "#/definitions/v30Package" },
+    { "$ref": "#/definitions/v40Package" }
```
jsancio: Let's give these definitions a different name. E.g. `v30BuildDef`

jsancio: Let's remove `v20Package` (or `v20BuildDef`) for now. I don't think the v2 version all all of the functionality we need.

repo/meta/schema/metadata-schema.json
```diff
@@ -0,0 +1,585 @@
...
+      "additionalProperties": false
+    }
+
+  },
+
+  "type": "object",
+  "oneOf": [
+    { "$ref": "#/definitions/v20Package" },
+    { "$ref": "#/definitions/v30Package" },
+    { "$ref": "#/definitions/v40Package" }
```
jsancio: Same as my comments in the other file.

Let's call these something else. E.g. `v40Metadata`.

Let's remove `v2` for now. It doesn't have all of the features we need.

PR #1343, opened 2017-08-10
Add a HDFS Non-HA DC/OS Service 

fmarchand:
As explained in this document [HDFS - Minimal installation](https://docs.mesosphere.com/service-docs/hdfs/v1.3.0-2.6.0-cdh5.9.1/install/#minimal-installation), the cost to run the HDFS service is prohibitive when you just want to provide a small HDFS cluster to few datascientists or just to test a simple program using HDFS. I needed a minimal installation without all the HA components (journal node, zkfc's and second namenode) because my cluster is small and I have other frameworks running on it.

I saw in this link that this contribution would be welcome so ... I wrote a HDFS Service based on the existing HDFS HA that requires less resources than this latter. I would like to contribute to the DC/OS Universe with this light version of HDFS.

All the files for this package (scheduler, bootstrap ...) are on S3. There's also a hadoop file which is the 2.8.1 version without all the documentation.   

Here is the [dcos-commons](https://github.com/fmarchand/dcos-commons) fork I did. I tried to write a README.md the most complete I could but maybe some informations are missing (that's my first PR to mesosphere universe). So don't hesitate to tell me if that's the case !


fmarchand:
Any update?

ryadav88:
@triclambert can you please take a look?

PR #1345, opened 2017-08-11
Transformer service

takirala:
Currently the Universe infrastructure pre-computes all of the possible Universes for the supported cluster when the Universe repository gets updated. This PR creates a service that will fetch a provided URL and will attempt to transform it to an Universe JSON that is understandable by the requesting DC/OS (Cosmos)

Addresses https://jira.mesosphere.com/browse/DCOS_OSS-1526


converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+        dcos_version
+    )
+    updated_json_data = json.dumps({json_key_packages: processed_packages})
+    errors = gen_universe.validate_repo_with_schema(
+        json.loads(updated_json_data),
+        repo_version
+    )
+    if len(errors) != 0:
+        logger.error(errors)
+        raise ValueError(ErrorResponse.VALIDATION_ERROR)
```
nickbp: Include the validation errors in the HTTP response?

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    repo_version = _get_repo_version(accept)
+    dcos_version = _get_dcos_version(user_agent)
+    logger.debug('Url %s\nAgent %s\nAccept %s\nVersion %s\nDC/OS %s',
+                 decoded_url, user_agent, accept, repo_version, dcos_version)
+
+    req = Request(decoded_url)
+    req.add_header(header_user_agent, user_agent)
+    req.add_header(header_accept, accept)
+    res = urlopen(req)
```
nickbp: This should specify a `timeout` value, some googling implies that the default could be no timeout

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    result = re.search(r'\bdcos/\b\d\.\d{1,2}', user_agent_header)
+    if result is None:
+        raise ValueError(ErrorResponse.UNABLE_PARSE)
+    return str(result.group().split('/')[1])
+
+
+class ErrorResponse:
+    INVALID_PATH = 'URL Path is invalid'
+    INVALID_HEADERS = 'Headers are invalid'
```
nickbp: This should say that the `Required 'User-Agent' and 'Content-Type' headers are missing`

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+def _get_repo_version(accept_header):
+    """Returns the version of the universe repo parsed.
+    # TODO Should support versions of two digits - y/n ?
+    :param accept_header: String
+    :return repo version as str or Error
+    :rtype str
+    """
+    result = re.findall(r'\bversion=v\d{1,2}', accept_header)
+    if result is None or len(result) is 0:
+        raise ValueError(ErrorResponse.UNABLE_PARSE)
```
nickbp: This and the error message at L160 should be different so that the caller can tell which header is the problem

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+    result = re.search(r'\bdcos/\b\d\.\d{1,2}', user_agent_header)
+    if result is None:
+        raise ValueError(ErrorResponse.UNABLE_PARSE)
+    return str(result.group().split('/')[1])
+
+
+class ErrorResponse:
+    INVALID_PATH = 'URL Path is invalid'
+    INVALID_HEADERS = 'Headers are invalid'
+    INVALID_PARAM = 'Request parameters are invalid'
```
nickbp: This should say that the `Required parameter 'url' is missing`

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+    :rtype str
+    """
+    result = re.search(r'\bdcos/\b\d\.\d{1,2}', user_agent_header)
+    if result is None:
+        raise ValueError(ErrorResponse.UNABLE_PARSE)
+    return str(result.group().split('/')[1])
+
+
+class ErrorResponse:
+    INVALID_PATH = 'URL Path is invalid'
```
nickbp: This should say that it `Expected path '/transform'`

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+    dcos_version = _get_dcos_version(user_agent)
+    logger.debug('Url %s\nAgent %s\nAccept %s\nVersion %s\nDC/OS %s',
+                 decoded_url, user_agent, accept, repo_version, dcos_version)
+
+    req = Request(decoded_url)
+    req.add_header(header_user_agent, user_agent)
+    req.add_header(header_accept, accept)
+    res = urlopen(req)
+    charset = res.info().get_param(param_charset) or default_charset
+    packages = json.loads(res.read().decode(charset)).get(json_key_packages)
```
nickbp: This should limit the number of bytes read to say 1MB. What if the destination URL is very large?

converter/service/converter.py
nickbp: How does this behave if the destination URL returns a 404?

converter/service/converter.py
nickbp: How does this behave if the content isn't JSON?

converter/service/converter.py
```diff
@@ -0,0 +1,175 @@
...
+from http import HTTPStatus
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from urllib.parse import parse_qsl, urlparse
+from urllib.request import Request, urlopen
+
+
+# Binds to all available interfaces
+HOST_NAME = ''
+# Gets the port number from $PORT0 environment variable
+PORT_NUMBER = int(os.environ['PORT0'])
```
nickbp: Should be able to use PORT_UNIVERSECONVERTER here (marathon also provides the port by name)

converter/service/converter.py
nickbp: Does this raise if the hostname isn't resolvable?

takirala:
@nickbp Below screenshot shows the following four cases 
- 404 Error 
- Unauthorized access
- Invalid json content and
- Hostname not found.

In all the cases, 400 is returned to user along with a corresponding error message (underlined in the screen capture). 
screen shot 2017-08-15 at 1 52 28 pm

For the size limit (number of bytes to read) what should be the expected behavior?

nickbp:
Maybe just cut losses and return an error if the limit is reached (as opposed to trying to parse the data you got)

takirala:
@jsancio @nickbp For the maximum content size, I believe 10 MB is a sensible default? I am testing with https://universe.mesosphere.com/repo and it returns content of ~7Mb

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+    dcos_version = _get_dcos_version(user_agent)
+    logger.debug('Version %s\nDC/OS %s', repo_version, dcos_version)
+
+    req = Request(decoded_url)
+    req.add_header(header_user_agent, user_agent)
+    req.add_header(header_accept, accept)
+    try:
+        res = urlopen(req, timeout=MAX_TIMEOUT)
+        charset = res.info().get_param(param_charset) or default_charset
+        raw_data = res.read()
```
nickbp: Maybe do a `read(MAX_BYTES+1)` here? Then the following check will still produce an error if the max is exceeded

scripts/validate_packages.py
```diff
@@ -40,7 +40,7 @@ def main():
                 "pattern : {}".format(letter, LETTER_PATTERN.pattern)
             )
         prefix_path = os.path.join(PKG_DIR, letter)
-        # traverse each package dir directory (ie "cassandra")
+        # traverse each package dir directory (e.g., "cassandra")
```
larioj: 😄

converter/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Set the working directory to the /converter directory
+WORKDIR /converter
+
+# Copy the required contents into the container at /converter
+ADD converter/service/converter.py          /converter
+ADD scripts/gen_universe.py                 /converter
+ADD repo/meta/schema                        /converter/repo/meta/schema
+ADD scripts/requirements/requirements.txt   /converter/requirements.txt
+
+# Install the required depen
```
jsancio: Spell this out. "dependencies".

converter/marathon.json
```diff
@@ -0,0 +1,36 @@
...
+    {
+      "name" : "universeconverter",
+      "port" : 8086,
+      "protocol" : "tcp"
+    }
+  ],
+  "labels": {
+    "DCOS_SERVICE_NAME": "transform",
+    "DCOS_SERVICE_PORT_INDEX": "0",
+    "DCOS_SERVICE_SCHEME": "http",
```
jsancio: We shouldn't expose this service through the admin router. Remove all of the `DCOS_SERVICE_*` labels.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from urllib.error import URLError, HTTPError
+from urllib.parse import parse_qsl, urlparse
+from urllib.request import Request, urlopen
+
+
+# Binds to all available interfaces
+HOST_NAME = ''
+# Gets the port number from $PORT0 environment variable
+PORT_NUMBER = int(os.environ['PORT_UNIVERSECONVERTER'])
```
jsancio: Where is the `PORT_UNIVERSECONVERTER` set? Do you mean `PORT0`?

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+class Handler(BaseHTTPRequestHandler):
+    def do_GET(s):
+        """
+        Respond to the GET request. The expected format of this request is:
+                http://<host>:<port>/transform?url=<url> with `User-Agent`
+                and `Accept` headers
+        """
+        errors = _validate_request(s)
+        if errors:
+            s.send_error(HTTPStatus.BAD_REQUEST, errors)
```
jsancio: You should use the `explain=` argument instead of override the HTTP response message string.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+            return
+
+        query = dict(parse_qsl(urlparse(s.path).query))
+        if param_url not in query:
+            s.send_error(HTTPStatus.BAD_REQUEST,
+                         ErrorResponse.PARAM_NOT_PRESENT.to_msg(param_url))
```
jsancio: You should use the `explain=` argument instead of override the HTTP response message string.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+        user_agent = s.headers.get(header_user_agent)
+        accept = s.headers.get(header_accept)
+        decoded_url = query.get(param_url)
+
+        try:
+            json_response = handle(decoded_url, user_agent, accept)
+            s.send_response(HTTPStatus.OK)
+            content_header = gen_universe.format_universe_repo_content_type(
+                _get_repo_version(accept))
+            s.send_header(header_content_type, content_header)
```
jsancio: We should send the `content-length`.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+            s.end_headers()
+            s.wfile.write(json_response.encode())
+        except Exception as e:
+            s.send_error(HTTPStatus.BAD_REQUEST, str(e))
```
jsancio: We shouldn't should do this here. This may execute after `s.send_response(HTTPStatus.OK)` was called and violate HTTP. We probably want:

```python
try:
  json_response = handle(decoded_url, user_agent, accept)
except ...:
  s.send_error(HTTPStatus.BAD_REQUEST, explain=...)
```
If we fail after that point then we just want to close the connection. How does `BaseHttpRequestHandler` and `HttpServer` handle an exception from `do_GET`?

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def handle(decoded_url, user_agent, accept):
+    """Results in either an error or the requested json data
```
jsancio: We don't return "error". We return a string (json encoding) or `raise` an exception, right? If so please update this comment.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+    logger.debug('Version %s\nDC/OS %s', repo_version, dcos_version)
+
+    req = Request(decoded_url)
+    req.add_header(header_user_agent, user_agent)
+    req.add_header(header_accept, accept)
+    try:
+        res = urlopen(req, timeout=MAX_TIMEOUT)
+        charset = res.info().get_param(param_charset) or default_charset
+        raw_data = res.read(MAX_BYTES+1)
+        if len(raw_data) > MAX_BYTES:
```
jsancio: We should just check the content-length of the response and fail if it doesn't have one or is greater than this value.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+
+    if header_accept not in s.headers:
+        return ErrorResponse.HEADER_NOT_PRESENT.to_msg(header_accept)
+
+
+def _get_repo_version(accept_header):
+    """Returns the version of the universe repo parsed.
+    # TODO Should support versions of two digits - y/n ?
+    :param accept_header: String
+    :return repo version as str or Error
```
jsancio: This only returns a string. It raises an Error.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+        raise ValueError(ErrorResponse.UNABLE_PARSE.to_msg(accept_header))
+    result.sort(reverse=True)
+    return str(result[0].split('=')[1])
+
+
+def _get_dcos_version(user_agent_header):
+    """Parses the version of dcos from the specified header.
+
+    :param user_agent_header: String
+    :return dcos version as str or Error
```
jsancio: This only returns a string. It raises an Error.

converter/service/converter.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+    updated_json_data = json.dumps({json_key_packages: processed_packages})
+    errors = gen_universe.validate_repo_with_schema(
+        json.loads(updated_json_data),
+        repo_version
+    )
+    if len(errors) != 0:
+        logger.error(errors)
+        raise ValueError(ErrorResponse.VALIDATION_ERROR.to_msg(errors))
+    return updated_json_data
```
jsancio: This needs to be simplified. At a high-level this:

1. Filter dict: dict -> dict
1. Project dict by enclosing it in a dict with `package` property: dict -> dict
1. Encode as a string using JSON: dct -> str
1. Decode string as a JSON into a dict: str -> dict
1. Validate the generated dict: dict -> errors

I think we can get rid of 3. and 4. if we return a `dict` or just 4. if we want to return a `str`.

converter/service/converter.py
takirala: so it seems that marathon provides the port with `PORT_<whatever name you specify in the marathon app json portDefinitions field>`. We can use `PORT0` or `PORT_UNIVERSECONVERTER`

converter/service/converter.py
takirala: @jsancio `Content-Length` right?

converter/service/converter.py
takirala: I modified the behavior. In case we throw an exception in do_GET this is what happens https://stackoverflow.com/questions/1924434/curl-error-52-empty-reply-from-server

converter/service/converter.py
```diff
@@ -109,12 +115,14 @@ def handle(decoded_url, user_agent, accept):
     try:
         res = urlopen(req, timeout=MAX_TIMEOUT)
```
jsancio: This returns a context manager so it needs to be closed: https://docs.python.org/3.6/glossary.html#term-context-manager

You can do this with:
```python
with urlopen(...) as res:
  ...
```

converter/service/converter.py
```diff
@@ -109,12 +115,14 @@ def handle(decoded_url, user_agent, accept):
...
-                         decoded_url,
-                         len(raw_data))
+
+        if header_content_length not in res.headers:
+            raise ValueError(ErrorResponse.ENDPOINT_HEADER_MISS.to_msg())
+
+        if int(res.headers.get(header_content_length)) > MAX_BYTES:
             raise ValueError(ErrorResponse.MAX_SIZE.to_msg())
+
+        raw_data = res.read(MAX_BYTES)
```
jsancio: We already checked the `content-length`. You can just do `raw_data = res.read()`.

PR #1350, opened 2017-08-15
[INFINITY-2196] Set the hello-world minDcosReleaseVersion to 1.8

elezar:
The minDcosReleaseVersion for hello-world was incorrectly set to 1.9 causing problems with deploying on 1.8 clusters. This PR corrects that.


nickbp:
The source value should also be updated: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/helloworld/universe/package.json#L3

elezar:
@nickbp the change to the source was done as part of mesosphere/dcos-commons#1398

PR #1355, opened 2017-08-18
Kong v0.11.0 release

shashiranjan84:
CHANGELOG: https://github.com/Mashape/kong/blob/master/CHANGELOG.md

Updates:
- SERF ports removed
- VIP added to port definition for internal traffic
- Migrations flag added

### DIFF
--- a/repo/packages/K/kong/3/config.json
+++ b/repo/packages/K/kong/3/config.json
@@ -1,0 +1,211 @@
+{
+  "$schema":"http://json-schema.org/schema#",
+  "type":"object",
+  "properties":{
+    "service":{
+      "type":"object",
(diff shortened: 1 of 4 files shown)


repo/packages/K/kong/4/config.json
```diff
@@ -0,0 +1,211 @@
+{
```
ryadav88: subdirectory version should be 3 not 4.

ryadav88:
@shashiranjan84 PR looks good apart from requested change above.

repo/packages/K/kong/4/config.json
shashiranjan84: @ryadav88 yes counting mistake. I have update it `3`. Please review

PR #1372, opened 2017-08-24
Update package.json

benclarkwood:


benclarkwood:
Un-break dcos-commons smoke

PR #1397, opened 2017-08-31
Add upgradesFrom and downgradesTo fields to the beta packages

gabrielhartmann:
Trial run with C* is already merged.


gabrielhartmann:
Install is fine on 1.10.

PR #1405, opened 2017-09-06
Bumped Scale package to 5.1.1

gisjedi:
Bump of Scale version to fix DCOS 1.10 support.  I did *not* add a new version to the Universe directory structure as this is just a patch release.  Let me know if that is not the correct way to bump for patches. Thanks!


ryadav88:
@gisjedi please create a new version directory

PR #1487, opened 2017-10-26
Relax url schema in build def to accept file paths as well

takirala:
We need this as we would use docker references and file paths. The universe packages are protected with the regular schema so we can relax this here.


larioj:
How did you validate that this works?

takirala:
@larioj Used https://www.jsonschemavalidator.net/ to check the file my CI generates

PR #1538, opened 2017-11-22
Traefik on DC/OS (DCOS_OSS-1807).

deric:
Traefik is a HTTP reverse proxy and load balancer that supports multiple backends (Marathon, etcd, consul, ZooKeeper, ...). Current version offers simple integration with Marathon backend(s). [Jira Ticket DCOS_OSS-1807](https://jira.mesosphere.com/browse/DCOS_OSS-1807).

### DIFF
--- a/repo/packages/T/traefik/0/config.json
+++ b/repo/packages/T/traefik/0/config.json
@@ -148,7 +148,7 @@
           "type": "boolean"
         },
         "expose": {
-          "default": true,
+          "default": false,
           "description": "Expose Marathon apps by default in Traefik.",
           "type": "boolean"
         },
--- a/repo/packages/T/traefik/0/config.json
+++ b/repo/packages/T/traefik/0/config.json
@@ -37,7 +37,7 @@
           "type": "string"
         },
         "minimumHealthCapacity": {
-          "default": 0.5,
+          "default": 0.0,
           "description": "Minimum health capacity.",
           "minimum": 0,
           "type": "number"
--- a/repo/packages/T/traefik/0/config.json
+++ b/repo/packages/T/traefik/0/config.json
@@ -9,15 +9,15 @@
           "type": "string"
         },
         "cpus": {
-          "default": 2,
+          "default": 1,
           "description": "CPU shares to allocate to each Traefik instance.",
           "minimum": 1,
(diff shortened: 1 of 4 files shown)


joerg84:
@deric This is great. PR lgtm, but @ryadav88 PTAL

bearnard:
Is there a way to specify something like maxDcosReleaseVersion? Not sure if this will work on 1.1x due to https://github.com/containous/traefik/issues/2115

deric:
@joerg84 Thanks Joerg. 
@bearnard Good point! Currently Traefik Marathon backend `go-marathon` is not fully compatible with Marathon 1.5 API gambol99/go-marathon#324. So, we would have support up to DC/OS 1.9. AFAIK there's no attribute `maxDcosReleaseVersion`. 

There's already a [PR in progress](https://github.com/gambol99/go-marathon/pull/327), hopefully `go-marathon` will be backward compatible with Marathon 1.4, otherwise we would have to require DC/OS >= 1.10. Maybe somebody from Marathon developers could verify the changes?

Meanwhile can we test backwards compatibility with earlier DC/OS releases?

ryadav88:
@deric  please add a documentation example for Traefik. You can create a PR against https://github.com/dcos/examples and use the template from other examples there. 
This is now a requirement for adding new packages to Universe.

deric:
@ryadav88 I've added basic documentation.
@bearnard IMHO missing support for `ipAddresses` feature is not a blocker right now. The upcoming Traefik 1.5.0 (a [rc1](https://github.com/containous/traefik/releases/tag/v1.5.0-rc1) is already available) might fix that. Routing to host ports or containers ports behind NAT works fine.

Traefik web interface is now accessible via DC/OS UI, also I've lowered default CPU and memory requirements. Even `cpu:0.1` and `mem:100` should be enough for testing Traefik.

GoelDeepak:
@deric is there a work in progress to support marathon 1.5?

deric:
@GoelDeepak Yes, see the comments above https://github.com/gambol99/go-marathon/pull/327
I'm testing Traefik against Maraton 1.5.1.2 (DC/OS 1.10.1) so far without problems, but I haven't tested VLANs yet.

Issue #1084, opened 2017-04-12
Support building local universe over an HTTP proxy


jcmcken:
Currently, some hacks are needed in order to build a local universe from behind a web filter/proxy (which is pretty standard in enterprise environments)

Here are some of the issues I've come across:

* ``docker/local-universe/Makefile`` - The ``local-universe`` target doesn't support passing ``HTTP_PROXY`` via environment. The bare ``sudo`` will not pass env vars into the Python process. If ``sudo`` environment passing is configured for the user, replacing ``sudo`` with ``sudo -E`` should do the trick.
* ``docker/local-universe/Makefile`` - The ``base`` target needs a ``--build-arg "http_proxy=$HTTP_PROXY"`` to pass the HTTP_PROXY env var into the container.

Other env vars might also need to be accounted for (e.g. ``HTTPS_PROXY``, ``NO_PROXY``, and the lowercase equivalents of these env vars).

jcmcken:
Related: the current workflow doesn't allow you to specify a directory for temporary files via the ``TMPDIR`` env var either. Since building the universe downloads a large number of artifacts, this can lead to an ``ENOSPC`` error (out of space on device) if your ``/tmp`` isn't large enough.

The ``sudo python3 scripts/local-universe.py`` would need to be ``sudo -E TMPDIR=$TMPDIR python3 ...`` (typical security configurations for ``sudo`` reset ``TMPDIR`` regardless of ``-E``, hence the explicit setting).

frankscholten:
I am also running into this issue.

joerg84:
@ryadav88 Could we address this issue or at least document it

ryadav88:
@jsancio @takirala ^

frankscholten:
Ready for review: https://github.com/mesosphere/universe/pull/1556

PR #1556, opened 2017-12-05
Added proxy build-args to the docker build command

frankscholten:
Fix for https://github.com/mesosphere/universe/issues/1084

### DIFF
--- a/docker/local-universe/Makefile
+++ b/docker/local-universe/Makefile
@@ -11,7 +11,7 @@ certs:
 		-subj "/CN=master.mesos"
 
 base: clean certs
-	docker build -t universe-base -f Dockerfile.base .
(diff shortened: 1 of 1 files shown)


jsancio:
Thanks @frankscholten! Do we need to change the documentation here https://github.com/mesosphere/universe/blob/version-3.x/docker/local-universe/README.md?

frankscholten:
@jsancio Perhaps add a section that mentions setting `http_proxy` and `https_proxy` when running behind a proxy?

jsancio:
> Perhaps add a section that mentions setting http_proxy and https_proxy when running behind a proxy?

Sounds good to me. I think we can add a sentence to bullet 1. in this section: https://github.com/mesosphere/universe/blob/version-3.x/docker/local-universe/README.md#building-your-own

PR #1558, opened 2017-12-06
Changed MLB healthcheck protocol to MESOS_HTTP from marathon HTTP

surdy:
Marathon has deprecated HTTP health checks and  recommended moving to MESOS_HTTP healthchecks https://github.com/mesosphere/marathon/releases/tag/v1.4.0

Just updated the latest version of the package 1.11.1 to use MESOS_HTTP instead of HTTP healtcheck

### DIFF
--- a/repo/packages/M/marathon-lb/31/config.json
+++ b/repo/packages/M/marathon-lb/31/config.json
@@ -1,0 +1,141 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "properties": {
+    "marathon-lb": {
+      "properties": {
+        "auto-assign-service-ports": {
+          "default": false,
+          "description": "Auto assign service ports for tasks which use IP-per-task. See https://github.com/mesosphere/marathon-lb#mesos-with-ip-per-task-support for details.",
+          "type": "boolean"
+        },
+        "bind-http-https": {
+          "default": true,
+          "description": "Reserve ports 80 and 443 for the LB. Use this if you intend to use virtual hosts.",
+          "type": "boolean"
+        },
+        "cpus": {
+          "default": 2,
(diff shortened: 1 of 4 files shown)


asridharan:
@surdy this is only for the Marathon-lb container. However, MLB should be using MESOS_HTTP healthchecks to ascertain the health check of the backends, as long as the backends enable MESOS_HTTP health checks. Need to see if that is the case.

Also @surdy I think we should cut a different release for MLB rather than just updating the latest package version defaults? @drewkerrigan ??

surdy:
@asridharan that's correct this was meant to be a change for just the MLB app in the universe not the changing the MLB's core behavior. We have seen customers and users run into issues with marathon HTTP healthchecks (https://github.com/mesosphere/marathon-lb/issues/504 and https://jira.mesosphere.com/browse/MARATHON-7572) and we have been recommending them to use MESOS_ healthchecks and since we should lead my example I changed this here. 

I was tempted to change the package version but then I noticed that MLB universe packages follow the released versions of MLB (https://github.com/mesosphere/marathon-lb/releases) and since the change in this case was outside of core MLB code there is no new version to associate this Universe package to.

drewkerrigan:
@asridharan I think it's fine to just create a new universe entry without an MLB release because this change only affects the configuration which only exists here in universe. I'm good with this PR.

drewkerrigan:
@surdy for the future we prefer these to be split into 2 commits - the first is copying the old version to a new dir, and the second commit containing the changes - so that it's easy to see exactly what changed.

asridharan:
Sounds good @drewkerrigan

surdy:
Closing this out since it did not get merged  and is already out of date

PR #1561, opened 2017-12-10
Fix escaping difference in cosmos

benclarkwood:
Prior to 1.10, Cosmos had a rendering bug (fixed by https://github.com/dcos/cosmos/commit/42c35b13527586daeba3359d1ed0602a59e7708d) that required
 stringified JSON to be doubly escaped. This was corrected
 in 1.10, but it means that packages with stringified JSON parameters
 that need to bridge versions must be accomodated.
```
 < 1.9 style escaping:
 \\\"field\\\": \\\"value\\\"

 >= 1.10 style escaping:
 \"field\": \"value\"
```

This uses a regex to detect \" and replace it with \\\" if version < 1.10. I can rewrite it as a parser if we're worried about having a regex.

### DIFF
--- a/scripts/gen_universe.py
+++ b/scripts/gen_universe.py
@@ -12,6 +12,7 @@ import pathlib
 import shutil
 import sys
 import tempfile
+import re
 import zipfile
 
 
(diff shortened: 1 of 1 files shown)


benclarkwood:
I tested this escaping manually by installing the universe converter in a cluster and then using that to install a package that works in both 1.10 and 1.9 but requires the extra escaping in 1.9.

jsancio:
@takirala Do you need to make any changes to the Universe transformer/converter to pick up this feature?

benclarkwood:
The converter calls filter_and_downgrade_packages_by_version directly, which is why I chose it as the injection point :).

benclarkwood:
Once we merge a fix, we will have to update the universe converter of couse.

takirala:
Ben is correct. I will update the converter once this is merged.

benclarkwood:
@takirala merged.

PR #1610, opened 2018-01-05
Gen Universe Tutorial

jsancio:


docs/gen_universe/gen_universe.markdown
```diff
@@ -0,0 +1,68 @@
...
+Clone the Universe repository with:
+```bash
+git clone git@github.com:mesosphere/universe
+cd universe
+```
+
+Assuming that a package was created and saved in the following location:
+```bash
+tree ~/work/private-universe/
+```
```
takirala: Why not merge both the code blocks OR add a text statement in between them if you feel user needs some help understanding.

jsancio: No good reason. One is a command and the other is the output. I wanted to make it easier for the user to copy and paste the command.

jsancio:
@jieyu I updated the documentation. Let me know if that works for you.

PR #1642, opened 2018-01-19
Jlee/add stub universe

justinrlee:
Add directions to inject stub universe contents (such as edge-lb repos) to a local universe.


docker/local-universe/add-stub-universe.sh
```diff
@@ -0,0 +1,45 @@
...
+for PACKAGE in $PACKAGES; do
+echo "Building repo structure for $PACKAGE..."
+LETTER=$(echo $PACKAGE | head -c 1 | tr '[:lower:]' '[:upper:]')
+mkdir -p stub-repo/packages/${LETTER}/${PACKAGE}/0
+
+jq --arg PACKAGE "$PACKAGE" '.packages[] | select(.name == $PACKAGE)' $FILE > stub-repo/$PACKAGE.json
+
+cat stub-repo/$PACKAGE.json | jq '.config'  > stub-repo/packages/${LETTER}/${PACKAGE}/0/config.json
+cat stub-repo/$PACKAGE.json | jq '.resource' > stub-repo/packages/${LETTER}/${PACKAGE}/0/resource.json
+cat stub-repo/$PACKAGE.json | jq '.command' > stub-repo/packages/${LETTER}/${PACKAGE}/0/command.json
```
jsancio: For the 3 commands above you can extract release value from `jq .releaseVersion`. What do you think? So it becomes `stub-repo/packages/${LETTER}/${PACKAGE}/${RELEASE}/....json`

benclarkwood:
I defer to @jeid64 as I know little about local universe.

benclarkwood:
We also have https://github.com/mesosphere/dcos-commons/blob/master/tools/release_builder.py which can be used for the same purpose more or less.

docker/local-universe/add-stub-universe.sh
justinrlee: Do we actually use releaseVersion?  Does it have to match the actual release version?  My concern is a little bit about what happens if somebody tries to load something with `.releaseVersion 2` and we haven't created a /0 and /1 version in the same repo.

justinrlee:
@benclarkwood I think release_builder.py is more for building our online universe (it pushes to S3, builds artifacts, etc.  This is more of a shim for injecting things into a local universe (basically, for airgapped customers who need Edge-LB, which has come up at least 3 or 4 times.

docker/local-universe/add-stub-universe.sh
justinrlee: I'm not enough of an expert on Universe to know if the it's more preferable to have:

Option 1:
* `/<correct-release-number>`, with no preceding release numbers

Option 2:
* `/0`, but not matching the `.releaseVersion` value

I mean, this script is simple enough that I'm okay with it not covering all use cases.  We just need to get *something* published so that customers in airgapped environments can use stub universe definitions without having to come to support (or me) every time.  Looking for an MVP at the moment.

jsancio: > My concern is a little bit about what happens if somebody tries to load something with .releaseVersion 2 and we haven't created a /0 and /1 version in the same repo.

This case should work. Sounds good to me to merge this change as is and improve it when the use case comes up.

PR #1650, opened 2018-01-22
Changed releaseVersion to temporarily remove Eremetic.

ssylvester87:
If Eremetic package is deployed under a Marathon deployment group, it can cause the mesos-master process to crash and take Marathon down with it. Temporarily removing package from the Universe until the edge case that causes this can be resolved.
See https://jira.mesosphere.com/browse/COPS-2254

### DIFF
--- a/repo/packages/E/eremetic/0/package.json
+++ b/repo/packages/E/eremetic/0/package.json
@@ -9,7 +9,7 @@
     }
   ],
   "maintainer" : "dcos-support@appliedis.com",
-  "minDcosReleaseVersion": "1.8",
+  "minDcosReleaseVersion": "999",
   "name": "eremetic",
   "packagingVersion": "3.0",
   "preInstallNotes": "This DC/OS Service is currently in preview. There may be bugs, incomplete features, incorrect documentation, or other discrepancies. Preview packages should never be used in production!",


jsancio:
Why temporarily? What needs to happens for us to add this package? Why not update the package so that it doesn't crash Marathon?

ssylvester87:
It's a community supported package, not one supported by Mesosphere. I've submitted COPS-2270 as well so that we can reach out to the appropriate owners to get the package fixed, but it's been suggested to get it removed from the Universe in the meantime.

surdy:
@jsancio The framework for the eremetic package registers with slashed (`/`) in the framework ID. Currently Mesos does not have checks around this and the slashed lead to mesos master failing. Mesos is going to add some validation to prevent master from failing in case of invalid framework ID https://issues.apache.org/jira/browse/MESOS-8470. But until them this package is a risk to the cluster causing master failures. 

Hence we want to "disable" this package until we have a version of Mesos with a check to avoid this issue. Unless @ryadav88 can work with the package committer to fix the framework to not use slashes in the framework ID

ryadav88:
@gisjedi is the original submitter. 
I've talked to him and created this issue: https://github.com/mesosphere/universe/issues/1684
Merging this PR for now. Jonathan will send a new PR soon.

PR #1656, opened 2018-01-28
New package for Zeppelin 0.7.3

swoehrl-mw:
New and reworked package for the current zeppelin version 0.7.3.
Based on my docker image from https://github.com/MaibornWolff/dcos-zeppelin

### DIFF
--- a/repo/packages/Z/zeppelin/6/resource.json
+++ b/repo/packages/Z/zeppelin/6/resource.json
@@ -7,7 +7,7 @@
   "assets": {
     "container": {
       "docker": {
-        "zeppelin": "maibornwolff/zeppelin:0.7.3-2-all"
+        "zeppelin": "maibornwolff/zeppelin:0.7.3-2-all-2.2.1-1-hadoop-2.6"
       }
     }
   }
--- a/repo/packages/Z/zeppelin/6/resource.json
+++ b/repo/packages/Z/zeppelin/6/resource.json
@@ -1,0 +1,14 @@
+{
+  "images": {
+    "icon-small": "https://downloads.mesosphere.com/zeppelin/assets/icon-service-zeppelin-small.png",
+    "icon-medium": "https://downloads.mesosphere.com/zeppelin/assets/icon-service-zeppelin-medium.png",
+    "icon-large": "https://downloads.mesosphere.com/zeppelin/assets/icon-service-zeppelin-large.png"
+  },
+  "assets": {
+    "container": {
+      "docker": {
+        "zeppelin": "maibornwolff/zeppelin:0.7.3-2-all"
+      }
+    }
+  }
+}
--- a/repo/packages/Z/zeppelin/6/config.json
+++ b/repo/packages/Z/zeppelin/6/config.json
@@ -43,12 +43,6 @@
         "spark": {
             "type": "object",
             "properties": {
-                "version": {
-                  "type": "string",
-                  "description": "The spark version to use",
-                  "enum": ["2.2.1-1-hadoop-2.6", "2.2.1-1-hadoop-2.7"],
-                  "default": "2.2.1-1-hadoop-2.6"
-                },
                 "cores_max": {
                     "type": "string",
                     "description": "Sets spark.cores.max",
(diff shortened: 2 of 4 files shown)


joerg84 (commented):
LGTM. I probably would have slight preference of 
a) having the docker image in the mesosphere/dcos dockerhub org
b)access to the dockerfile for repeatable builds.

@ryadav88 @ArtRand @susanxhuynh Could you take another look?

ryadav88:
@joerg84 For Zeppelin package updates, [Tweeter demo](https://github.com/mesosphere/tweeter) should be tested.   
https://jira.mesosphere.com/browse/DCOS_COMMUNITY-90 cc: @pleia2

swoehrl-mw:
@joerg84 The dockerfile is contained in the repo https://github.com/MaibornWolff/dcos-zeppelin. If you want, you can take the code into the mesosphere github org and let it push images into the mesosphere dockerhub org.
@ryadav88 Unfortunately I will not have time to test the tweeter demo in the next few days. So would be great if somebody else could run it.

cc: @pleia2

repo/packages/Z/zeppelin/6/resource.json
```diff
@@ -0,0 +1,14 @@
+{
+  "images": {
+    "icon-small": "https://downloads.mesosphere.com/zeppelin/assets/icon-service-zeppelin-small.png",
+    "icon-medium": "https://downloads.mesosphere.com/zeppelin/assets/icon-service-zeppelin-medium.png",
+    "icon-large": "https://downloads.mesosphere.com/zeppelin/assets/icon-service-zeppelin-large.png"
+  },
+  "assets": {
+    "container": {
+      "docker": {
+        "zeppelin": "maibornwolff/zeppelin:0.7.3-2-all"
```
takirala: @swoehrl-mw  Is this image publicly accessible? (Not sure if I am looking at the wrong page but i cant see any image with this tag at https://hub.docker.com/u/maibornwolff/)

swoehrl-mw: @takirala The image is publicly available from https://hub.docker.com/r/maibornwolff/zeppelin/

susanxhuynh (commented):
I didn't have time to test it, but looks good. Thanks.

repo/packages/Z/zeppelin/6/resource.json
takirala: @swoehrl-mw i could see the image but the tag is not visible (or is incorrect). I caught this because an internal CI failed when trying to build this docker image. 

These are all the tags that I can see : 

screen shot 2018-02-06 at 11 38 31 am

docker pull also fails : 
```
> docker --debug pull maibornwolff/zeppelin:0.7.3-2-all
Error response from daemon: manifest for maibornwolff/zeppelin:0.7.3-2-all not found
```
Am I missing something ?

swoehrl-mw: @takirala If you check the marathon.json.mustache you can see that the image is defined as "{{resource.assets.container.docker.zeppelin}}-{{spark.version}}". So depending on the spark version selected in the options another image is used as I want to provide images with different spark versions (at the moment only 2.2.1-hadoop-2.6/2.7 as it is the current version).
I agree that it is a little confusing that the image name defined in the resources is not the full image/tag. Do you know of a better way to handle this?

takirala: @ryadav88 FYI This would fail the package.dcos build CI. @jsancio Is it okay to ignore the CI build failure for this case ? we won't be able to generate a package.dcos for zepplin.

jsancio: @swoehrl-mw The problem with this solution is that we can't build `package.dcos` or Local Universe. This means that users won't be to run zeppelin in an air gapped cluster.

One possible solution is to enumerate all of the images that you need in `{{resource.assets.container.docker...}}` and picking the right one based on the version of spark. Another solution is to have a different package for each version of Spark just like we do with Cassandra, Kafka, etc.

jsancio (dismissed):
See my other comment.

swoehrl-mw:
@jsancio I implemented your second suggestion. The idea is then to build/release a new package version each time spark or zeppelin get a new version. For the moment I only provide the package for zeppelin 0.7.3 and spark 2.2.1 (with hadoop 2.6, as this is the version the official dc/os hdfs package uses).

joerg84:
Tested with tweeter (old functioning version).

joerg84 (approved):

PR #1688, opened 2018-02-01
Eremetic framework fix

### LINKED ISSUE #1684: Installing "Eremetic" universe package resulted in dcos-mesos-master and dcos-marathon to enter failed state
User reports that after installing the community package Eremetic, the mesos-master and marathon services entered a failed state on all masters.

The framework for the eremetic package registers with slashed (/) in the framework ID. Currently Mesos does not have checks around this and the slashed lead to mesos master failing. Mesos is going to add some validation to prevent master from failing in case of invalid framework ID https://issues.apache.org/jira/browse/MESOS-8470. But until them this package is a risk to the cluster causing master failures.

Hence we want to "disable" this package until we have a version of Mesos with a check to avoid this issue.

Please fix the framework to not use slashes in the framework ID

gisjedi:
#1684: Resolves issue of framework id being specified with a '/' in the name. Prohibits specifying / in the `framework-name` config.json field.

### DIFF
--- a/repo/packages/E/eremetic/0/config.json
+++ b/repo/packages/E/eremetic/0/config.json
@@ -9,7 +9,8 @@
         "framework-name": {
           "description": "The name of the Eremetic scheduler and service instance",
           "type": "string",
-          "default": "eremetic"
+          "default": "eremetic",
+          "pattern": "^[a-zA-Z0-9-]+$"
         },
         "cpus": {
           "description": "Number of cores allocated to Eremetic instance",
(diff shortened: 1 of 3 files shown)


surdy:
Thanks @gisjedi 

@ryadav88 I did not realize that it was not default package but the users input that lead to bad framework ID. At any rate good to have this validation. Do we have external facing documentation on guiding package contributors ? If yes then it would be a good to document this there.

ryadav88:
@surdy we have a general tutorial but not for best practices. https://github.com/mesosphere/universe/blob/version-3.x/docs/tutorial/GetStarted.md

PR #1690, opened 2018-02-02
Corrected base64 decode switch

aggress:
Confirmed on CentOS and Ubuntu


aggress:
Thought so. This can't move forwards as is without checking for OS type. -D is correct for darwin but not for Linux. I will add that check.

docker/local-universe/add-stub-universe.sh
```diff
@@ -47,14 +54,14 @@ jq --arg PACKAGE "$PACKAGE" '.packages[] | select(.name == $PACKAGE)' $FILE > st
...
+cat stub-repo/$PACKAGE.json | jq -r '.marathon.v2AppMustacheTemplate' | ${BASE64} > stub-repo/packages/${LETTER}/${PACKAGE}/0/marathon.json.mustache
 
 done
 
 echo ""
 echo "Full stub-repo contents:"
 ls -alh stub-repo/packages/*/*/*
 
-echo ""
+echo ""
```
jsancio: Add a newline.

aggress: He's on 🌴  so may not get to see this for a bit

jsancio: Sounds good. Merge it when you are ready.

PR #1724, opened 2018-03-02
missed packaged that currently dont have a image object

fatz:
as dcos-enterprise-cli and dcos-aws-cli etc. did not had an image until now the script was simply ignoring them.

Here is the fix. 

However on https://mesosphere.com/service-catalog/all are some other packages with image errors. Currently its unclear why. Investigating


jsancio (approved):
LGTM.

PR #1777, opened 2018-04-05
Add service icon instructions and files

ashenden:


ashenden:
@jsancio is there a folder in this directory where we could host the two template files linked in my README update seen here?  Currently I'm hosting on CloudApp, but would rather host and version track it here in Github.

ashenden:
@fatz this PR adds two template files to help package authors create icons.  If you have any specific instructions about how to get those files hosted by us, could you add as well?

fatz:
So I see two options:

1. have a raw-image attribute in the resource json and we use e.g. convert to the our expected formats
2. have a CI job searching for an icon.png in this git repo and then do convert and upload

So IMO the second option has the benefit of not being depending on any other remote service. So everything could be done with just pulling the universe git. However this also will blew up the size of this git repo.

ashenden:
@fatz ultimately I defer to you guys on this.  Both options make sense to me. My concern with the first option is we'd have to place the Base64 content for all 3 images in each resource.json, which means we don't have the option of being selective on the frontend about what image we choose to pull.

fatz:
@ashenden I don't think it has to be a base64. So just placing the raw image in the repo should be enough... then we could upload the raw image to that bucket. The different versions can be generated with a simple CI script. 

So every time the raw image is change we just regenerate the smaller versions

ashenden:
ping @jsancio

PR #1778, opened 2018-04-06
Justinrlee fix crdb

justinrlee:
I'm pretty sure items in the resource.json should be absolute links, not relative paths, cause we use to build local universe (i.e., we try to actually download everything referenced in the resource.json)

### DIFF
--- a/repo/packages/C/cockroachdb/3/marathon.json.mustache
+++ b/repo/packages/C/cockroachdb/3/marathon.json.mustache
@@ -26,7 +26,7 @@
     "FRAMEWORK_NAME": "{{service.name}}",
     "FRAMEWORK_USER": "{{service.user}}",
     "FRAMEWORK_PRINCIPAL": "{{service.service_account}}",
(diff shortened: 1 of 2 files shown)


ryadav88:
cc: @a-robinson

a-robinson (approved):
If this is how it's done, you guys certainly know better than me.

I guess this means that we'll effectively have to hard-code the version in [our pre-build `resource.json` file](https://github.com/cockroachdb/dcos-cockroachdb-service/blob/master/universe/resource.json#L9)?

justinrlee:
Hey @a-robinson yeah, it's a limitation of the way we do package versions.  Typically, for a new release of a package (CockroachDB, in this case), we'd set up a DC/OS package release (for example, a full set of package.json, resource.json, etc. at `repo/packages/C/cockroachdb/4/` (note the 4 instead of 3).

This item in particular is causing issues for our local universe package mechanism, which is used to pull in resources into a local repository for clusters without access to the Internet.  We try to download and re-host everything listed in resource.json, which doesn't work with the relative linking thing in the initial version of this.

PR #1870, opened 2018-05-28
Add single package repo files for DC/OS >= 1.8

takirala:
Address [DCOS-37596](https://jira.mesosphere.com/browse/DCOS-37596)

- I have implemented this for DC/OS 1.8 and above.
- An example url would be `http://.../package/cassandra`. It should be straightforward to append `?version=<a.b.c>` in future if needed.
- Tested this by making sure i can add two repos with urls <domain-name>/repo and <domain-name>/package/cassandra to `Package Repositories`.

TODO 
- Need to coordinate with @lloesche once this is merged for the deployment of universe server.

### DIFF
--- a/docker/server/build.bash
+++ b/docker/server/build.bash
@@ -19,7 +19,7 @@ function clean {
 
 }
 
-function prepare {
+function prepare {(
 
   local universeBuildTarget="${DOCKER_SERVER_DIR}/../../target"
 
@@ -29,12 +29,14 @@ function prepare {
     cp -r ${DOCKER_SERVER_DIR}/../../target/repo-*.json ${DOCKER_SERVER_DIR}/target
     # copy over the build zip repos (only 1.6.1 and 1.7)
     cp -r ${DOCKER_SERVER_DIR}/../../target/repo-*.zip ${DOCKER_SERVER_DIR}/target
+    # copy over the packages files
+    (cd ${DOCKER_SERVER_DIR}/../../target; rsync -R */package/* ${DOCKER_SERVER_DIR}/target)
   else
     err "Please run scripts/build.sh before trying to build universe server"
   fi
 
 
-}
+)}
 
 function gzipJsonFiles {(
 
--- a/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
+++ b/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
@@ -62,6 +62,11 @@ server {
     rewrite ^/repo-1.7$ /repo-up-to-1.7.zip break;
   }
 
+  location /package {
+    # This is supported from 1.8 and above. Results in 404 for older versions.
+    rewrite ^ /$universe_version/$dcos_release_version$uri.json;
+  }
+
   location = /repo {
     types {
       application/zip zip;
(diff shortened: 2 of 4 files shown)


docker/server/build.bash
```diff
@@ -29,12 +29,14 @@ function prepare {
     cp -r ${DOCKER_SERVER_DIR}/../../target/repo-*.json ${DOCKER_SERVER_DIR}/target
     # copy over the build zip repos (only 1.6.1 and 1.7)
     cp -r ${DOCKER_SERVER_DIR}/../../target/repo-*.zip ${DOCKER_SERVER_DIR}/target
+    # copy over the packages files
+    (cd ${DOCKER_SERVER_DIR}/../../target; rsync -R */package/* ${DOCKER_SERVER_DIR}/target)
```
benclarkwood, line 33: Does this significantly increase the footprint of the service?

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -78,20 +83,22 @@ server {
     rewrite ^/repo$ /repo-up-to-$dcos_release_version.zip break;
   }
 
-  location = /v3 {
+  location /v3 {
```
benclarkwood, line 86: For my own edification, `=` meant "exact match"

scripts/gen_universe.py
```diff
@@ -705,6 +709,39 @@ def validate_repo_with_schema(repo_json_data, repo_version):
     return errors
 
 
+def _populate_dcos_version_json_to_folder(dcos_version, outdir):
+    """Populate the repo-up-to-<dcos-version>.json file to a folder.
+    The folder structure would be :
+        <dcos-version>
+            -package
+                -<name-of-package1>.json
```
benclarkwood: No need to change this now, but I would use `/` when demonstrating directory layouts. The `-`'s were confusing 

e.g.
```
<dcos-version>/
  package/
    <name1>.json
    <name2>.json
    ...
```

scripts/gen_universe.py
```diff
@@ -705,6 +709,39 @@ def validate_repo_with_schema(repo_json_data, repo_version):
...
+    :param outdir: Path to the directory to use to store all universe objects
+    :type outdir: str
+    :return: None
+    """
+    repo_dir = outdir / dcos_version / 'package'
+    pathlib.Path(repo_dir).mkdir(parents=True)
+    repo_file = pathlib.Path(outdir / 'repo-up-to-{}.json'.format(dcos_version))
+    with repo_file.open('r',  encoding='utf-8') as f:
+        data = json.loads(f.read())
+        packages_dict = {}
```
benclarkwood, line 731: Map of package name -> array of package definitions

scripts/gen_universe.py
```diff
@@ -705,6 +709,39 @@ def validate_repo_with_schema(repo_json_data, repo_version):
...
+
+        for package in data.get('packages'):
+            package_name = package.get('name')
+            package_list = packages_dict.get(package_name, [])
+            package_list.append(package)
+            packages_dict[package_name] = package_list
+
+        for package_name, package_list in packages_dict.items():
```
benclarkwood, line 739: Is there a reason for this separation? Seems like you could just do everything in a single for loop, no?

benclarkwood (approved):
Looks good overall.

Couple of comments, but I'm ok with it as is.

docker/server/build.bash
takirala: Yep, I believe so, we are adding a single file for each package for each DC/OS version. I see no harm though.

scripts/gen_universe.py
takirala: Yes and No.
I don't think i can do in single for loop. In the first for loop I am iterating over each package and gathering the definitions. This has to be completely done before I write each package (with all its gathered definitions) to a single json file in the second for loop.
However, as all the packets are currently organized in a sequential fashion, we can rely on that to parse all the packages and dump them to json whenever a new package is encountered. But i think its more safe to not rely on the fact that the packages are sorted by name. (And even more so, two for loops are better than sorting the array!)

benclarkwood: Fair enough

benclarkwood (approved):

PR #1871, opened 2018-05-30
bump marathon 1.6.490

meln1k:


ichernetsky (approved):

kensipe (changes requested):
this is a bugging version for which @timcharper is working on a solution.  It needs to be updated with the fix or rolled back.

meln1k:
The bug is fixed in 1.6.490

kensipe (approved):

timcharper (approved):

PR #1873, opened 2018-06-04
Remove json-schema.org from configs

nickbp:
Let's see how this does...

`find . -iname *.json | xargs sed -i '/json-schema.org/d'`

### DIFF
--- a/repo/meta/schema/build-definition-schema.json
+++ b/repo/meta/schema/build-definition-schema.json
@@ -1,5 +1,4 @@
 {
-  "$schema": "http://json-schema.org/draft-04/schema#",
 
   "definitions": {
 
@@ -198,7 +197,6 @@
 
 
     "config": {
-      "$ref": "http://json-schema.org/draft-04/schema#"
     },
 
 
--- a/repo/meta/schema/config-schema.json
+++ b/repo/meta/schema/config-schema.json
@@ -1,6 +1,4 @@
 {
-    "id": "http://json-schema.org/draft-04/schema#",
-    "$schema": "http://json-schema.org/draft-04/schema#",
     "description": "Core schema meta-schema",
     "definitions": {
         "schemaArray": {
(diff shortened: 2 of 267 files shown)


repo/meta/schema/build-definition-schema.json
```diff
@@ -198,7 +197,6 @@
 
 
     "config": {
-      "$ref": "http://json-schema.org/draft-04/schema#"
```
elezar, line 201: Is an empty `config` object allowed?

repo/meta/schema/config-schema.json
```diff
@@ -1,6 +1,4 @@
 {
-    "id": "http://json-schema.org/draft-04/schema#",
```
elezar, line 2: This is not a reference to a schema, but a field in the JSON. Are there any other implications to changing this?

repo/meta/schema/v3-repo-schema.json
```diff
@@ -253,7 +252,6 @@
 
 
     "config": {
-      "$ref": "http://json-schema.org/draft-04/schema#"
```
elezar, line 256: Another empty `config` object.

takirala (approved):
LGTM

elezar (approved):
Let's hope this works @nickbp.

It may be good to link a ticket or description of the reason for this change.

There are a couple of places where the replacements are "non-standard" and I don't have enough context to know if these may cause problems.

If they are OK, then this looks good from my side.

As a note, we don't specify the schema in any of our own SDK-based packages.

repo/meta/schema/build-definition-schema.json
nickbp: Seems to be... It's not clear why the schema needed to be referenced again like this, given that it's already referenced at the root level.

repo/meta/schema/config-schema.json
nickbp: Looks like someone messed up. This field is supposed to be a pointer to _your_ schema URL:

https://spacetelescope.github.io/understanding-json-schema/basics.html
```
Declaring a unique identifier
It is also best practice to include an id property as a unique identifier for each schema. For now, just set it to a URL at a domain you control, for example:

{ "id": "http://yourdomain.com/schemas/myschema.json" }
```

PR #1883, opened 2018-06-19
Change Configuration for collector

nexclouding:
Correct typo from collctor to collector


ssipflow:
We just correct typo for collector.

PR #1896, opened 2018-07-06
Divide Nexcloud into NexCloud-collector, NexCloud-workflow, NexCloud

nexclouding:
We just divided NexCloud into 3 components


nexclouding:
We divide our package NexCloud into three components (NexCloud-collector, NexCloud-workflow, NexCloud)

PR #1906, opened 2018-07-12
Print JSON snippet when error encountered

deric:
Better error message when JSON validation fails.

### DIFF
--- a/scripts/local-universe.py
+++ b/scripts/local-universe.py
@@ -278,8 +278,18 @@ def selected_check(package_json, packages, only_selected, is_latest):
 
 
 def load_json(json_path):
-    with json_path.open(encoding='utf-8') as json_file:
-        return json.load(json_file)
+    try:
+        with json_path.open(encoding='utf-8') as json_file:
+            return json.load(json_file)
+    except json.JSONDecodeError as err:
+        print("JSON error in file: %s" % json_path)
+        print_width = 50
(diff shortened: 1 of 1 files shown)


scripts/local-universe.py
```diff
@@ -278,8 +278,19 @@ def selected_check(package_json, packages, only_selected, is_latest):
 
 
 def load_json(json_path):
-    with json_path.open(encoding='utf-8') as json_file:
-        return json.load(json_file)
+    try:
+        with json_path.open(encoding='utf-8') as json_file:
+            return json.load(json_file)
+    except json.JSONDecodeError as err:
```
takirala, line 284: How about : 
```
except json.JSONDecodeError as err:
    print("JSON error in file: %s" % json_path)
    print_width = 50
    snippet = err.doc[max(0, err.pos - print_width) : err.pos + print_width]
    if err.pos - print_width > 0:
        snippet = '... {}'.format(snippet)
    if err.pos + print_width < len(err.doc):
        snippet = '{} ...'.format(snippet)
    print('{}\n{}'.format(err, snippet))
```
- This doesn't have temporary variables `start, stop`.
- The appending/prepending of `...` is consistent in style
- Just use one random number value of `50` instead of `20, 40, 60`. Makes the code more readable & easily editable if someone wants to print more content.

deric: @takirala Sure, I agree it's more readable. I've rebased and updated the file to your version.

takirala:
Thanks @deric

PR #1918, opened 2018-07-18
Remove support@mesosphere.io as maintainer for packages not supported by Mesosphere

surdy:
I was just going through the universe repo to check any packages that claim to be supported by Mesopshere but are not actually supported/ never were supported.
I found a few. Thankfully they are older versions of a bunch community packages. This PR updates package.json for all such old packages

### DIFF
--- a/repo/packages/Z/zeppelin/0/package.json
+++ b/repo/packages/Z/zeppelin/0/package.json
@@ -3,7 +3,7 @@
   "name": "zeppelin",
   "version": "0.5.6",
   "scm": "https://github.com/apache/incubator-zeppelin",
-  "maintainer": "support@mesosphere.io",
+  "maintainer": "https://dcos.io/community/",
   "description": "Zeppelin is a web-based notebook that enables interactive data analytics",
   "framework": true,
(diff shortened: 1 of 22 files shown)


ryadav88 (approved):

repo/packages/Z/zeppelin/0/package.json
```diff
@@ -3,7 +3,7 @@
   "name": "zeppelin",
```
ryadav88, line 3: @surdy unrelated to your changes, but `packagingVersion` 2.0 is not supported anymore and the PR tests fail on that. Could you update that for `Zeppelin` packages ?

ryadav88 (dismissed):

repo/packages/Z/zeppelin/0/package.json
surdy: If that's going to block this PR then I would suggest either 
1) Removing the packages from repo altogether
2) I can modify this PR to not touch the 2.0 packages and the owner of the Universe repo can update the packages that are still using 2.0

ryadav88: lets do `2`

PR #1945, opened 2018-08-08
Updated CEREBRO_VERSION string to 0.7.2

a-nldisr:
When this is set incorrect the cerebro package will fail to launch because this string is evaluated by the bootstrap.sh script and will cause a failure because it checks for a previous version:

```
I0807 12:40:08.788480     7 executor.cpp:651] Forked command at 12
sh: ./cerebro-0.7.1/bin/cerebro: No such file or directory
I0807 12:40:08.872782     8 executor.cpp:938] Command exited with status 127 (pid: 12)
```


deric:
Duplicate of #1782. @ryadav88 can you have a look at this?

ryadav88:
Fixed in https://github.com/mesosphere/universe/pull/1782

PR #1961, opened 2018-08-23
Add V5 packaging schema

takirala:
Notes

- We are NOT supporting v5 in local universe. (worth noting that we have package registry now instead of local universe, but its an EE only feature). If this is really needed, we can do this in future.


kvish:
Looks good to me but can we make universe permissive if it seens a header it does not understand? Following our current model we will always go cosmos feature with a new repo/package -> universe changes to understand that header/schema -> cosmos integration tests that can hit the universe + cosmos universe client updates. If we  make universe permissive we can commit the cosmos feature/tests together then make the universe change. It can be very possible we forget to add the new version to the universe client in cosmos

takirala:
As discussed offline and after doing some testing universe is already permissive @kvish

kvish:
Can we make the same change in converter than?

kvish (approved):

nickbp (approved):

takirala:
Thanks for the review @kvish @nickbp

PR #1966, opened 2018-08-30
Add zipkin package

DanielMSchmidt:
Hey there,

I found that I use zipkin quite a lot, so I thought I might just add it. Zipkin is a tool for performance tracing.

I have a little problem with the logos, they are cut. What can I do if I don't find them in the right size and what is the size they should have?

This is not urgent, so please don't rush with the review.

Thank you for your time 👍

### DIFF
--- a/repo/packages/Z/zipkin/0/config.json
+++ b/repo/packages/Z/zipkin/0/config.json
@@ -1,0 +1,84 @@
+{
+  "type": "object",
+  "properties": {
+    "service": {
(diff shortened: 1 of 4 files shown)


deric:
@ryadav88 Could you have a look at this?

deric:
@DanielMSchmidt Please create a PR with simple example describing the package into examples repository: https://github.com/dcos/examples

PR #1970, opened 2018-08-31
Remove leading tabs from package.json *InstallNotes / Add *InstallNotes to README

juliangieseke:
While fixing a bug in DC/OS UI, I recognized that there are quite some tabs in markdown fields of packages package.json

This PR is basically two questions: 
1. is it intended to have markdown in these fields? we're parsing them as such in the UI already and some packages also do it already. Also it would be a nice feature.
2. is it necessary to have the tabs for something? if not and we can remove them, we can also remove the hotfix in our UI and allow code blocks :tada: (intending something with a tab in markdown means there is a code block)

if both questions can be answered with `yes`, I'd love to get this merged 😇
if its a `no` - I'd love to know the reasons and if we can make "it", whatever reason it might be, markdown compatible 👍


juliangieseke:
test failed with `write /var/lib/docker/tmp/GetImageBlob418773441: no space left on device` 🤔

takirala:
The tabs are used to print the content as indented in CLI which renders it as plain text and pipes it to stdout.

```
By Deploying, you agree to the Terms and Conditions https://mesosphere.com/catalog-terms-conditions/#certified-services
Default configuration requires 3 agent nodes each with: 0.5 CPU | 4096 MB MEM | 1 10240 MB Disk
Continue installing? [yes/no] yes
Installing Marathon app for package [cassandra] version [2.3.0-3.0.16]
Installing CLI subcommand for package [cassandra] version [2.3.0-3.0.16]
New command available: dcos cassandra
The DC/OS Apache Cassandra service is being installed!

	Documentation: https://docs.mesosphere.com/service-docs/cassandra/
	Issues: https://docs.mesosphere.com/support/
```

PR #1987, opened 2018-09-18
set beta-kubernetes and beta-mke packages to selected=true

MatApple:


takirala:
This resulted in failing CI. `beta-mke` and `beta-kubernetes` are not selected in Universe Checker.

takirala:
@MatApple we usually don't mark `beta-*` packages as selected. Is there a reason for making these beta packages selected? If so, it needs to be updated here https://github.com/mesosphere/universe-checker/blob/master/checker-rules/src/test/scala/com/mesosphere/universe/checker/UniverseVerificationSpec.scala#L33

MatApple:
Oh really, ok so I'll revert.

PR #2013, opened 2018-10-10
Spinnaker release 0.3.0-1.9.2

realmbgl:


ryadav88 (approved):

ryadav88:
ignoring .dcos test failure per @takirala

takirala:
@realmbgl @ryadav88 is there a reason for not mentioning the `tag` for the `redis` docker image in the `resource.json` ? We should not assume `latest` tag because this would alter the image contents based on **when** the docker pull happens. And for building `.dcos` files, we download the docker image and then calculate the SHAsum for the file. Using a moving tag (such as `latest`) may cause some inconsistencies for this.

Is there a strong reason for not using a specific `tag` for the `redis` image ?

realmbgl:
@takirala , thx for your comment. We tag it to a specific version when we respin to support spinnaker 1.10, which will be soon. redis never caused any trouble.    cc @ryadav88

PR #2025, opened 2018-10-22
universe: remove old kubernetes packages

bmcustodio:
This PR removes `beta-kubernetes`, `beta-mke` and old versions of `kubernetes` from the universe.


MatApple:
Closing because we do not want to do this. Customers and field still rely upon these services and it may be some time before they migrate their workloads to a 2.0 Kubernetes cluster.

PR #2045, opened 2018-11-14
Catch DC/OS 1.13 user agents in Docker Nginx config

bamarni:


bamarni:
@takirala : could you confirm this? I think it's missing in the main branch

takirala (approved):

takirala:
Thanks @bamarni

PR #2074, opened 2018-12-03
Catch DC/OS 1.13 user agents in Docker Nginx config

bamarni:
This adds support for DC/OS 1.13 in docker nginx configurations.

### DIFF
--- a/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
+++ b/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
@@ -53,6 +53,10 @@ server {
     set $dcos_release_version 1.12;
   }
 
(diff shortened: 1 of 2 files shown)


docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -53,6 +53,10 @@ server {
     set $dcos_release_version 1.12;
   }
 
+  if ($http_user_agent ~ ".*dcos\/1\.13.*") {
```
takirala, line 56: This is for local universe and is enforced above in Line 34. 
For prod universe server, this is enforced at https://github.com/mesosphere/prod-universe/blob/master/nginx/etc/nginx/conf.d/universe.mesosphere.com.conf#L44

takirala:
Merging this for consistency.

PR #2080, opened 2018-12-03
Fix CockroachDB licenses

mattj-io:


ryadav88 (changes requested):
@mattj-io it needs a `,` delimiter after each license.

mattj-io:
@ryadav88 fixed and repushed

PR #2082, opened 2018-12-03
Fix dcos-core-cli licenses

mattj-io:


ryadav88 (approved):

bamarni:
@mattj-io : just saw this PR, I assume this is something we have to add from now on? There have been a few versions added in the meanwhile which don't have this.

mattj-io:
@bamarni yes we need to have that license - we'll do a last pass over universe at the end of this project and pick up any stragglers.

mattj-io:
@bamarni do we need a fix for the upstream package as well ?

PR #2085, opened 2018-12-03
Fix kubernetes license

mattj-io:


ryadav88 (approved):

MatApple (commented):
This package is our own package and doesn't have base tech. Do we want to resolve to this license? https://raw.githubusercontent.com/kubernetes/kubernetes/master/LICENSE

The `kubernetes-cluster` package has the kubernetes base tech.

mattj-io:
@MatApple - what's actually in this package ? Is this the framework code ? If so, that's proprietary and this PR isn't correct ..

MatApple:
@mattj-io this package and our `kubernetes-cluster` packages are proprietary. This package contains only code we've written.

`kubernetes-cluster` package contains kubernetes code and code we've written to run it on DC/OS.

The code in both these packages is closed source.

mattj-io:
@MatApple we still need to acknowledge the Apache licensed code for the portion we're redistributing. In the context of this exercise for Universe, it probably means listing both Apache and our proprietary license for the kubernetes-cluster package.

MatApple:
That makes sense 👍 Thanks @mattj-io. I'll approve.

MatApple (approved):

mattj-io:
I've added both Apache and proprietary to all of these packages, since I think kubernetes-cluster is pulled in by kubernetes, meaning it's deploying Apache code even if it doesn't contain it directly.

ryadav88 (approved):

PR #2087, opened 2018-12-03
Fix portworx licenses

mattj-io:


mattj-io:
@ryadav88 fixed and repushed

ryadav88:
cc: @disrani-px

PR #2088, opened 2018-12-03
Fix tensorflow licenses

mattj-io:


ryadav88 (approved):
cc @vishnu2kmohan, @disrani-px  so we can add a license in framework repo as well.

Issue #2109, opened 2018-12-19
Grafana 5.5.0-5.1.3 fails to deploy in strict mode.


jdijt:
When installing the grafana 5.5.0-5.1.3 package in strict mode (dcos 1.12.0) the installation fails with the following message:
```
$ dcos package install --options=grafana-options.json grafana
By Deploying, you agree to the Terms and Conditions https://mesosphere.com/catalog-terms-conditions/#community-services
Continue installing? [yes/no] yes
Unable to parse the string as a JSON value
```
using the following grafana-options.json:
```
{
    "service":{
        "user": "nobody",
        "service_account": "grafana-principal",
        "service_account_secret": "grafana/sa_key"
    }
}
```

PR #2110, opened 2018-12-19
#2109 Fix embedded json escaping for Grafana 5.5.0-5.1.3

jdijt:
Fix for issue #2109

We were able to figure out it was a parsing error of the marathon Mustache template, the fix is to escape the bit of embedded json in MESOS_MODULES. (As it is also done for other packages like, like Prometheus, see: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/P/prometheus/1/marathon.json.mustache#L56 )


jdijt:
Hmm, any guidance on how to prevent these checks from failing?

I executed build.sh locally without error, and was able to successfully build this version into a local universe for testing the new grafana template.

vishnu2kmohan:
@jdijt Only Mesosphere employees can trigger the TeamCity tests, which I've done now on your behalf.

vishnu2kmohan (approved):
Thanks @jdijt!

PR #2017, opened 2018-10-12
Support GPU scheduling behavior (MARATHON-8088,MARATHON-8089)

deric:
Add support for [GPU scheduling behavior](https://github.com/mesosphere/marathon/blob/master/changelog.md#gpu-scheduling). Supported from Marathon 1.6.352.

JIRA Issues:
 * https://jira.mesosphere.com/browse/MARATHON-8088
 * https://jira.mesosphere.com/browse/MARATHON-8089

```
diff --git a/202/config.json b/203/config.json
index 72d7d811..6b2ff8df 100644
--- a/202/config.json
+++ b/203/config.json
@@ -141,6 +141,16 @@
           "description" : "(Default: service.name) Framework name to register with Mesos.",
           "type" : "string"
         },
+        "gpu-scheduling" : {
+          "description" : "Define how offered GPU resource should be treated.",
+          "type" : "string",
+          "default" : "undefined",
+          "enum": [
+            "undefined",
+            "restricted",
+            "unrestricted"
+          ]
+        },
         "ha" : {
           "description" : "Enabled: (Default) Run Marathon in HA mode with leader election. Allows starting an arbitrary number of other Marathons but all need to be started in HA mode. This mode requires a running ZooKeeper Disabled: Run Marathon in single node mode.",
           "type" : "boolean",
diff --git a/202/marathon.json.mustache b/203/marathon.json.mustache
index b5261e1f..be941201 100644
--- a/202/marathon.json.mustache
+++ b/203/marathon.json.mustache
@@ -36,7 +36,7 @@
   "env": {
     "JVM_OPTS": "-Xms{{jvm.heap-min}}m -Xmx{{jvm.heap-max}}m {{#jvm.opts}}{{jvm.opts}}{{/jvm.opts}}"
   },
[...]

PR #2120, opened 2019-01-07
Fix GPU scheduling mustache

alenkacz:
#2017 introduced gpu-scheduling config parameter but that was not added correctly to the cmd so it ended up with following error `Failed to close \'marathon.gpu-scheduling\'`

This PR fixes that behavior

JIRA: DCOS-46647


zen-dog (approved):
Thx!

jeschkies (approved):
I trust that it works.

PR #2213, opened 2019-03-11
Revert to env based secrets to fix Jenkins on strict mode

kvish:


takirala (approved):
LGTM modulo testing manually in strict mode (in `minDcosReleaseVersion` if applicable)

PR #2330, opened 2019-07-08
chore: add new dcos-ui package version

juliangieseke:
This updates ui packages devliered for ui-update-service to version that we shipped with 1.13.2 for customers running 1.13.1 or 1.13.0.

### DIFF
--- a/repo/packages/D/dcos-ui/2/package.json
+++ b/repo/packages/D/dcos-ui/2/package.json
@@ -1,0 +1,22 @@
+{
+  "packagingVersion": "4.0",
+  "name": "dcos-ui",
+  "version": "2.82.6",
+  "tags": [
+    "dcos-ui",
+    "dcos",
+    "ui"
+  ],
+  "maintainer": "frontend-dev@mesosphere.io",
+  "description": "DC/OS UI",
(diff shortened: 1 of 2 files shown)


repo/packages/D/dcos-ui/2/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "packagingVersion": "4.0",
+  "name": "dcos-ui",
+  "version": "2.82.6+0e5782c0",
```
TattdCodeMonkey: we don't want a sha in this version.

TattdCodeMonkey (changes requested):

repo/packages/D/dcos-ui/2/package.json
TattdCodeMonkey: at least we should test this.

juliangieseke: CI also complains, yeah… not sure why tho, its valid semver 🤔

TattdCodeMonkey (approved):
Tested this with a stub universe, DC/OS 1.13.1 and `dcos-ui-update-service` and everything was 💯

PR #2335, opened 2019-07-17
Fixed CEREBRO_VERSION. jre tar contains a jdk folder, fixed that.

a-nldisr:
CEREBRO_VERSION was still pointing to old version, raised this. 
Also the JRE tar contains a jdk folder, changed the CMD to use these objects.


a-nldisr:
Any update?

PR #2341, opened 2019-07-31
add configuration option in schema registry to bootstrap servers

zmalik:
`SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS` is not optional for schema registry when configured TLS/SSL.

This PR adds the configuration option `kafka.kafkastore-bootstrap-servers` to let users configure it correctly.


jrx (approved):
Thank you @zmalik! Looks good to me and should be sufficient!

PR #2344, opened 2019-08-02
fix the parsing error in confluent schema registry

zmalik:
fixing the wrong addition of the config option `kafka.kafkastore-bootstrap-servers`


jrx (approved):
LGTM!

PR #2378, opened 2019-09-10
[COPS-5339] Fix resource URIs and make Marathon app template use them. 

mpereira:
The Marathon app template was using values relative to artifact-dir instead of
just using the resource URIs present in resources.json, which were
incidentally, incorrect.

This PR fixes the resource URIs and makes the Marathon app use them. This is the
root cause for the recent weirdness we've experienced with the latest Kibana
version Universe package.

This was fixed in the [dcos-elastic-service repository](https://github.com/mesosphere/dcos-elastic-service/pull/42) and I'm propagating the
changes here manually.

### DIFF
--- a/repo/packages/K/kibana/700/marathon.json.mustache
+++ b/repo/packages/K/kibana/700/marathon.json.mustache
@@ -62,13 +62,13 @@
       "cache": true
     },
     {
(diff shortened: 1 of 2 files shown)


jrx (approved):
LGTM

PR #2387, opened 2019-09-23
Add knownIssues and lastModified fields for v3, v4 and v4 package schemas

takirala:
Jira : DCOS-58438

### DIFF
--- a/repo/meta/schema/vX-repo-definitions.json
+++ b/repo/meta/schema/vX-repo-definitions.json
@@ -454,6 +454,14 @@
         "$ref": "#/dcosReleaseVersion",
         "description": "The minimum DC/OS Release Version the package can run on."
       },
+      "knownIssues": {
+        "type": "boolean",
+        "description": "Boolean field indicating if the package has any known issues."
+      },
+      "lastUpdated": {
+        "type": "number",
+        "description": "A unix epoch depicting the last updated date of the package base tech"
+      },
       "marathon": {
         "$ref": "#/marathon"
       },
@@ -576,6 +584,14 @@
       "minDcosReleaseVersion": {
         "$ref": "#/dcosReleaseVersion",
         "description": "The minimum DC/OS Release Version the package can run on."
+      },
+      "knownIssues": {
+        "type": "boolean",
+        "description": "Boolean field indicating if the package has any known issues."
+      },
+      "lastUpdated": {
+        "type": "number",
+        "description": "A unix epoch depicting the last updated date of the package base tech"
       },
       "marathon": {
         "$ref": "#/marathon"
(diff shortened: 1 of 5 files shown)


repo/meta/schema/vX-repo-definitions.json
```diff
@@ -577,6 +585,14 @@
         "$ref": "#/dcosReleaseVersion",
         "description": "The minimum DC/OS Release Version the package can run on."
       },
+      "knownIssues": {
```
ryadav88: I don't think `boolean` would fit the need here. If there are known issues, do we use something else to point users to what those issues are? Can we use string arrays?

repo/meta/schema/vX-repo-definitions.json
```diff
@@ -577,6 +585,14 @@
...
+        "type": "boolean",
+        "description": "Boolean field indicating if the package has any known issues."
+      },
+      "lastUpdated": {
```
ryadav88, line 592: how is this timestamp calculated?

repo/meta/schema/vX-repo-definitions.json
takirala: Sure. And what do we put in these string arrays ?

repo/meta/schema/vX-repo-definitions.json
takirala: @porridge has a ticket to update the release builder script in dcos-commons to calculate and embed the timestamp when making the universe PR. But I guess this is applicable for ceritifed packages only. If we want to add this field to ALL the packages, then we can probably modify the current CI to auto add this value when generating the universe file. Do you think that would be valuable ?

repo/meta/schema/vX-repo-definitions.json
samvantran: sorry about that @takirala, the original field we wanted to work off was `hasKnownIssues` which can totally be a boolean and is much less ambiguous. Can we update to use that field? See earlier tickets:
- https://jira.mesosphere.com/browse/DCOS-58017
- https://jira.mesosphere.com/browse/DCOS-58030

Btw, all we're trying to signify to users is this package has issues with deploying cleanly with default settings. We don't need to list what those issues are. The maintainers of the package should make sure that the package can be
1) deployed successfully with default configs or
2) if additional params are needed, they need to be clear about them in the description and make them required params

repo/meta/schema/vX-repo-definitions.json
samvantran: We need to upgrade all packages with this field regardless if they're certified or not.
> we can probably modify the current CI to auto add this value when generating the universe file

this would be great. We should also have CI in universe that checks for this field and if it's not present it throws an error so all community package contributors are required to add this field.

kvish:
lgtm from a code perspective

kvish (approved):

PR #2391, opened 2019-09-27
Add last modified field when generating universe json files

takirala:
This PR brings the following changes
- Add support for auto addition of `lastUpdated` field (if not already present).
- Stop updating zip files for DC/OS versions 1.6.1 and 1.7 (Previous URLs would continue to work - we just stop updating them with new packages)
- Rename 1.14 to 2.0

### DIFF
--- a/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
+++ b/docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
@@ -6,8 +6,6 @@ server {
 
   gzip_static on; # nginx will automatically search for a .gz version of a file and serve it if it exists
 
-  set $dcos_release_version 1.6.1;
-
   set $serve_json false;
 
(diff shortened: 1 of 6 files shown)


takirala:
FYI @ryadav88 @bamarni @porridge this PR adds couple fields and also defaults to 2.0 instead of 1.14 as the latest DC/OS version.

docker/server/nginx/etc/nginx/conf.d/universe.marathon.mesos.conf
```diff
@@ -58,7 +58,7 @@ server {
   }
 
   if ($http_user_agent ~ ".*dcos\/1\.14.*") {
```
kvish: change regex to dcos 2.0

kvish (approved):

kaiwalyajoshi (approved):

PR #2368, opened 2019-08-30
[DCOS-58017] Mark error-prone packages with `hasKnownIssues: true` 

rathohit311356:
## What changes were proposed in this pull request?
PR updates all the error-prone packages with the new field "hasKnownIssues". 
This is to mark all packages that do not deploy successfully with default values.
Resolves [Mark error-prone packages with `hasKnownIssues: true` ](https://jira.mesosphere.com/browse/DCOS-58017)

PR #2393, opened 2019-09-29
DCOS-58016 Update all universe packages with time of last update

ishanchandrajoshi:
This branch contains the updated package.json files with last_updated field.


takirala (dismissed):
This PR is superseded by https://github.com/mesosphere/universe/pull/2391
Merging this would affect the commit refs and thus this PR should not be merged

samvantran:
@takirala the PR you referenced (#2391) is for all future package builds but we need something that updates all the _current_ packages w/ their `lastUpdated` time (this PR). 

PR #2391 also looks like it updates the field if it doesn't already exist (see [here](https://github.com/mesosphere/universe/pull/2391/files#diff-c4ca3bb11e5c023a1b98e58d2c58dcf6R290)). Doesn't this PR help to add that field? 

One thing this PR needs to fix is that CI fails for all package schema versions before 3.0. However, for those packages - we still need to relay information that they are extremely outdated. How should we proceed with that? @ishanchandrajoshi see PR #2387 as `lastUpdated` only works from schema v3+ 

cc @pierrebeitz for packages v2 and older, look like we won't have relevant `lastUpdated` fields - perhaps the UI can default to showing a warning? Something like `The last time this package was updated could not be determined indicating this package is likely very outdated`? WDYT?

takirala:
> @takirala the PR you referenced (#2391) is for all future package builds but we need something that updates all the current packages w/ their lastUpdated time (this PR).

I am not sure I got this right, #2391 updates ALL (not just future) packages with `lastUpdated` field (if it is not already present) as long as it is v3 or above. We "refresh" our  universe repo on every merge to default branch.

samvantran:
@takirala I see. So since #2391 is the latest commit if we merge one additional PR all the packages in this repo will immediately get updated with the new field?

Isn't that a little heavy handed? Committing to the default branch w/o any reviews? Seems a bit dangerous.

takirala:
https://github.com/mesosphere/universe/pull/2391/files#diff-c4ca3bb11e5c023a1b98e58d2c58dcf6R285-R296 is the part which is responsible for adding the `lastUpdated` field for each package based on the commit for the specific package folder. This may be unique for each package folder (for e.g.: https://github.com/mesosphere/universe/tree/version-3.x/repo/packages/C/cassandra/0 differs from https://github.com/mesosphere/universe/tree/version-3.x/repo/packages/C/cassandra/300 )

samvantran:
Not sure if you're intentionally dodging my question. 

I'll ask again: in the next commit to the default branch `version-3.x`, all package.jsons in this github repository will have the new field `lastUpdated` added, so long as their package version is 3+. Is this correct?

If so, can you make a simple commit to the default branch to trigger the "refresh" so that we have the new field? PR #2368 is waiting on this update so that it doesn't mess with the commit time of any folder.

takirala:
Ah may be i conveyed this incorrectly. no one is pushing commits to any branch with `lastUpdated` field. #2391 added tooling to append this `lastUpdated` metadata when it is generating the universe repo json files. This will never be part of package.json but will be part of the generated repo json file (Like https://downloads.mesosphere.com/universe/repo/repo-up-to-2.0.json) and is calculated dynamically on every merge to universe default branch.

HOWEVER, consider all of this as stale because we cannot really go with this approach as airgapped solutions (local universe - technically deprecated but still being used by customers so we have to support) does not work with this approach. So I am reverting the change via #2398 and we can go ahead with this PR. We can revisit this if/when local universe is completely deprecated but I guess we have to go with this PR for now. Sorry for the confusion. We can chat offline if this is still unclear.

samvantran:
Got it - okay thanks for clarifying. Yup, let's continue with this PR. Basically we need to update every package with the time it was last updated and we're following a similar approach: the time the folder was committed to the repo. That way, whether using local universe or not, the UI will be able to pick up this information and display to users how recent (or outdated) the package is.

ishanchandrajoshi:
Closing this pull request as a new pull request is raised #2401 for updating Universe packages with lastUpdated field considering schema version.

PR #2398, opened 2019-10-02
Do not populate lastUpdated field in package.json : Revert #2391 partially

takirala:
The previous approach of populating metadata inside package.json (`lastUpdated` specifically) does not work for airgapped solutions ([COPS-5402](https://jira.mesosphere.com/browse/COPS-5402?focusedCommentId=297040&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-297040)). This PR reverts the universe tooling changes that were adding the field in package.json

We should rely on #2393 to add this field. Sorry for the mess @porridge @samvantran and @kvish

### DIFF
--- a/scripts/gen_universe.py
+++ b/scripts/gen_universe.py
@@ -5,7 +5,6 @@ import argparse
 import base64
 import collections
 import copy
-import git
 import itertools
 import json
 import jsonschema
@@ -17,8 +16,8 @@ import re
 import zipfile
 
 
-project_dir = "{}/..".format(os.path.dirname(os.path.realpath(__file__)))
-schema_dir = os.environ.get("SCHEMA_DIR", "{}/repo/meta/schema/".format(project_dir))
+dir_path = os.path.dirname(os.path.realpath(__file__))
+schema_dir = os.environ.get("SCHEMA_DIR", "{}/../repo/meta/schema/".format(dir_path))
 repo_definitions_json = "{}/vX-repo-definitions.json".format(schema_dir)
 
 
@@ -53,10 +52,6 @@ def main():
         print('The path in --repository [{}] is not a directory.'.format(
             args.repository))
         return
-    else:
-        global git_repo
-        git_repo = git.Git(project_dir)
-        print("universe git repo detected at {}".format(project_dir))    
 
     packages = [
         generate_package_from_path(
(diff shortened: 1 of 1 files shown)


scripts/gen_universe.py
```diff
@@ -282,18 +277,9 @@ def read_package(path):
...
 
-    package_json_path = path / 'package.json'
-
-    with package_json_path.open(encoding='utf-8') as file_object:
-        package_json = json.load(file_object)
-
-        # Update the `lastUpdated` field in the `package.json` if it is not already present.
-        if (LooseVersion(package_json.get('packagingVersion', '0.0')) >= LooseVersion("3.0")) \
-                and ("lastUpdated" not in package_json):
-            last_updated = git_repo.log("-1", '--date=unix', '--format=%cd', '--', path)
```
kvish, line 293: I still like the idea of getting the last updated time from the git repo where possible. Why not catch the exception and continue or check for the presence of a .git directory? we can still support the git repo approach for some cases

samvantran: +1 it would be good to still have the tooling to populate this value for SDK FWs to benefit future releases

takirala: I thought the same too, but, the fact that this does NOT work all the time means there would be some ambiguity and we need to have a backup anyway. I feel not touching the `package.json` at all would be more ideal and feels like WYSIWYG. For certified packages, we would rely on dcos-commons tooling to inject the field so IMO that automation is good enough for now( i believe Marcin had a ticket for that)

takirala: Also, this tooling is not sufficient for packages published at https://downloads.mesosphere.com/universe/packages/packages.html - which would be weird cuz then a package can have a lastUpdated in public universe but NOT in airgapped environment and I thought reverting the change would be better to be consistent.

porridge: I don't really know how this local-universe build thing is supposed to work (is it the customer who runs it, or us, or both?), but either way we're missing some cases - unless updating the SDK tooling will be enough to make such packages as `dcos-ui` or `marathon` expose `lastUpdated`?

cneth: The local universe build is run by customers who need a customized, local version of the Universe that only contains the packages they need.   It is also possible to add your own site-local or customized packages using this procedure.   At any rate, since the current state of the master is broken, I'd like to suggest we merge this change and then have a separate discussion about how to best implement the 'lastUpdated' field.

samvantran: +1 if this change resolves COPS-5402 then let's merge and get back to a working state

kvish (approved):

scripts/gen_universe.py
cneth: Thanks!

PR #2401, opened 2019-10-04
[DCOS-58016] Add lastUpdated field to Universe packages with schema version v3, v4

ishanchandrajoshi:
This branch contains the modified package.json for all universe packages having schema version v3, v4.

### DIFF
--- a/repo/meta/schema/build-definition-schema.json
+++ b/repo/meta/schema/build-definition-schema.json
@@ -304,6 +304,14 @@
           "$ref": "#/definitions/dcosReleaseVersion",
           "description": "The minimum DC/OS Release Version the package can run on."
         },
+        "hasKnownIssues": {
+          "type": "boolean",
+          "description": "Boolean field indicating if the package has any known issues."
+        },
+        "lastUpdated": {
+          "type": "number",
+          "description": "A unix epoch depicting the last updated date of the package base tech"
+        },
         "marathon": {
           "$ref": "#/definitions/marathon"
         },
@@ -421,6 +429,14 @@
           "$ref": "#/definitions/dcosReleaseVersion",
           "description": "The minimum DC/OS Release Version the package can run on."
         },
+        "hasKnownIssues": {
+          "type": "boolean",
+          "description": "Boolean field indicating if the package has any known issues."
+        },
+        "lastUpdated": {
+          "type": "number",
+          "description": "A unix epoch depicting the last updated date of the package base tech"
+        },
         "marathon": {
           "$ref": "#/definitions/marathon"
         },
@@ -531,6 +547,14 @@
           "required": [
             "packageName"
           ]
+        },
+        "hasKnownIssues": {
+          "type": "boolean",
+          "description": "Boolean field indicating if the package has any known issues."
+        },
+        "lastUpdated": {
+          "type": "number",
+          "description": "A unix epoch depicting the last updated date of the package base tech"
         },
         "marathon": {
           "$ref": "#/definitions/marathon"


docs/tutorial/time-server/0/package.json
```diff
@@ -4,5 +4,10 @@
...
   "maintainer": "https://github.com/mesosphere/universe",
   "description": "This is a simple python http server that displays a webpage which says the current time at the server location.",
-  "tags": ["http", "time-server", "basichttp"]
-}
+  "tags": [
+    "http",
+    "time-server",
+    "basichttp"
+  ],
+  "lastUpdated": 1499380201000
```
takirala: How did you calculate this value? This is not unix epoch.

takirala: FYI I have used https://github.com/mesosphere/universe/pull/2398/files#diff-c4ca3bb11e5c023a1b98e58d2c58dcf6L293 to calculate the epoch from commit ref

repo/packages/A/airfield/0/package.json
```diff
@@ -1,25 +1,26 @@
...
+  "postInstallNotes": "Airfield is being installed!\n\n\tDocumentation: https://github.com/MaibornWolff/dcos-airfield\n\tIssues: https://github.com/MaibornWolff/dcos-airfield",
+  "postUninstallNotes": "Airfield is being uninstalled.",
+  "licenses": [
+    {
+      "name": "Apache License 2.0",
+      "url": "https://raw.githubusercontent.com/MaibornWolff/dcos-airfield/master/LICENSE"
+    }
+  ],
+  "lastUpdated": 1552664069000
+}
```
takirala: (nit) can you ensure no newline characters are trimmed from files.

takirala: feel free to add a newline at the end AND retain the newline if it is already present - applicable for ALL files in this PR.

takirala:
Also can you do the same change for v5 as well. I _think_ mke and k8 cluster are the only v5 packages for now. 
Thanks :)

docs/tutorial/time-server/0/package.json
ishanchandrajoshi: It is the Unix time epoch in milliseconds. On running `git log --format=%aD docs/tutorial/time-server/0/package.json` we get _Thu, 6 Jul 2017 15:30:01 -0700_. On converting the aforementioned timestamp to actual date we get _Thursday, July 6, 2017 10:30:01 PM_ in UTC which is equal to above time.

repo/packages/A/airfield/0/package.json
ishanchandrajoshi: > Also can you do the same change for v5 as well. I _think_ mke and k8 cluster are the only v5 packages for now.
> Thanks :)

I think v5 packages should have been included in this change. I wrote a logic which would update all packages having version greater than v2.

docs/tutorial/time-server/0/package.json
takirala: Okay, please change this value to seconds.

ishanchandrajoshi: Alright

ishanchandrajoshi: Done

samvantran (approved):
Nice! Thanks @ishanchandrajoshi. This will help a lot with letting users know how active packages are.

takirala (approved):
**This is a nit**

Can you adjust your script to not make any whitespace changes (such as changing a tab (or four spaces) to 2 spaces). I am all in for standardising a 2 level indent across the universe repo but this PR is not helping that case because after merging this PR in current state a single pacakge would have `package.json` with 2 level indent and its other files with a different indent which kind of makes it worse. But, more importantly, merging this PR would result in less "useful" git blame (as we would be touching all the irrelevant lines). It would make me very happy if you adjust your scripting to not inject any whitespaces. Ideally, this PR should have #files = #linesofdiff (one line per file). 

But this is not a blocker and just a (easy) nice to have :)

ishanchandrajoshi:
> **This is a nit**
> 
> Can you adjust your script to not make any whitespace changes (such as changing a tab (or four spaces) to 2 spaces). I am all in for standardising a 2 level indent across the universe repo but this PR is not helping that case because after merging this PR in current state a single pacakge would have `package.json` with 2 level indent and its other files with a different indent which kind of makes it worse. But, more importantly, merging this PR would result in less "useful" git blame (as we would be touching all the irrelevant lines). It would make me very happy if you adjust your scripting to not inject any whitespaces. Ideally, this PR should have #files = #linesofdiff (one line per file).
> 
> But this is not a blocker and just a (easy) nice to have :)

Hi @takirala, I have changed the identation from 2 spaces to four spaces.

PR #2408, opened 2019-10-11
Enables Connector classes in Confluent Replicator

wjglerum:
An extra environment variable is needed to enable the Connector classes, see https://github.com/confluentinc/cp-docker-images/issues/427


zmalik (approved):
LGTM thank you @wjglerum !

PR #2414, opened 2019-10-16
Add lastUpdated field for universe packages.

takirala:
Inherited from  #2401 Thanks @ishanchandrajoshi 

This is the script created with help from @ishanchandrajoshi  : 
```python
import json
import subprocess
import os
import os.path
from datetime import datetime


def get_last_updated(filepath):
    out = subprocess.run('git log --format=%aD {} | head -1'.format(filepath), shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    time = datetime.strptime(str(out.stdout.decode("utf-8")[:-1]), '%a, %d %b %Y %H:%M:%S %z')
    last_updated_time = (time.timestamp())
    return int(last_updated_time)


def get_all_package_paths(basedir):
    paths = []
    for dirpath, dirnames, filenames in os.walk(basedir):
        for filename in [f for f in filenames if f.endswith("package.json")]:
            paths.append(os.path.join(dirpath, filename))
    return paths


def run_cmd_fetch_stdout(cmd):
    # print("executing {}".format(cmd))
    out = subprocess.run(cmd, shell = True, stdout = subprocess.PIPE, stderr = subprocess.PIPE)
    result = out.stdout.decode("utf-8").rstrip()
    # print("return {}".format(result))
    return result


def main():
    paths = get_all_package_paths(os.getcwd())
    remaining = len(paths)
    for path in paths:
        with open(path, "r+") as f:
            package = json.load(f)
            if float(package.get("packagingVersion")) > 2.0:
                last_updated = get_last_updated(path)
                lines = int(run_cmd_fetch_stdout("echo $(wc -l < {})".format(path)))
                run_cmd_fetch_stdout("gsed -i -e '{}s/$/,/' {}".format(lines - 1, path))
                indent = int(run_cmd_fetch_stdout("detect-indent {} | wc -c".format(path)))
                run_cmd_fetch_stdout("""gsed -i '{}i\{}"lastUpdated": {}' {}""".format(lines, indent * " ", last_updated, path))
                remaining = remaining - 1
                if remaining % 10 == 0:
                    print("updated {}, remaining {}".format(path, remaining))
            else:
                remaining = remaining - 1


if __name__ == "__main__":
    main()
```


kaiwalyajoshi (approved):

bamarni:
@takirala : does this field has to be set manually when adding/updating packages? Or are we going to periodically make PRs like this one to set it on all packages?

takirala:
@bamarni yes that is correct.

PR #2418, opened 2019-10-18
[DCOS-58017] Mark error-prone packages with WARNING

rathohit311356:
## What changes were proposed in this pull request?
PR updates all the error-prone packages with the Warnings under pre-install notes and additional of new field hasknownissues 
This is to mark all packages that do not deploy successfully with default values.
Resolves [Mark error-prone packages with `hasKnownIssues: true` ](https://jira.mesosphere.com/browse/DCOS-58017)


takirala:
I do not understand why the hasKnownIssues field is not used. What was the point of adding that field if we don’t use it in packages?

samvantran:
@rathohit311356, when I said to use the pre-install notes as a place to put the warning, I meant only for packages whose package version was v2 or older. For all packages v3 and above, we have the `hasKnownIssues` field.

@takirala for context, Hitesh was getting validation errors with the `hasKnownIssues` field for packages like `concord` which is on version v2. But my suggestion was only meant for those super outdated packages.

rathohit311356:
@samvantran @takirala  My bad , i have corrected it now 
Could you please re-review

pierrebeitz (approved):

orlandohohmeier (approved):

PR #2424, opened 2019-10-22
Use downloads.mesosphere.io since Github releases aren't working with Cosmos

janisz:
Github return 403 on HEAD requests that's why we need to switch to our mesosphere.downloads.io.

Refs:
* https://gitlab.com/janisz/dcos-http-cli/-/tags/0.1.0
* https://twitter.com/janiszt/status/1186300349956550656

### DIFF
--- a/repo/packages/H/helloworld/0/resource.json
+++ b/repo/packages/H/helloworld/0/resource.json
@@ -6,11 +6,11 @@
           "contentHash": [
             {
               "algo": "sha256",
-              "value": "f0028105a0bea830d5bdc9f07a3f9c9c3af433a016a6bf1b7921cf646e397376"
+              "value": "7f06eb113794b9e021fb396b65c18c9f2b426a6c2443c0444e359c4be7a6dc89"
             }
           ],
           "kind": "zip",
-          "url": "https://github.com/dcos/dcos-http-cli/releases/download/0.1.0/dcos-http-cli.darwin.zip"
+          "url": "https://gitlab.com/janisz/dcos-http-cli/uploads/36c122df16f0a43015d766ddeb27ba7d/dcos-http-cli.darwin.zip"
         }
       },
       "linux": {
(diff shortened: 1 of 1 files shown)


repo/packages/H/helloworld/0/resource.json
```diff
@@ -6,35 +6,35 @@
           "contentHash": [
             {
               "algo": "sha256",
-              "value": "f0028105a0bea830d5bdc9f07a3f9c9c3af433a016a6bf1b7921cf646e397376"
+              "value": "7f06eb113794b9e021fb396b65c18c9f2b426a6c2443c0444e359c4be7a6dc89"
             }
           ],
           "kind": "zip",
-          "url": "https://github.com/dcos/dcos-http-cli/releases/download/0.1.0/dcos-http-cli.darwin.zip"
+          "url": "https://gitlab.com/janisz/dcos-http-cli/uploads/36c122df16f0a43015d766ddeb27ba7d/dcos-http-cli.darwin.zip"
```
makkes: We really shouldn't reference private repos here.

makkes (dismissed):

repo/packages/H/helloworld/0/resource.json
janisz: I agree but I need a place where I can put this files. We can upload it to S3 but I don't know who to ask for it or create team on gitlab.

drewvanstone: Looks like it might be the ROQ team: https://wiki.mesosphere.com/display/ENG/ROQ+-+Publishing+to+downloads.mesosphere.com

janisz: Thanks for link. It's working

downloads.mesosphere.io/dcos-http-cli/0.1.0/dcos-http-cli.darwin.zip
downloads.mesosphere.io/dcos-http-cli/0.1.0/dcos-http-cli.linux.zip
downloads.mesosphere.io/dcos-http-cli/0.1.0/dcos-http-cli.windows.zip

janisz:
@takirala Can you fix docker build process?

repo/packages/H/helloworld/0/resource.json
drewvanstone: Awesome! Thanks for switching it over

bamarni (commented):
My only concern here is regarding the mismatch between the CLI plugin and package name, we've seen some tests failing because of this.

As the helloworld package in its current form is just a file server:
 https://github.com/mesosphere/universe/blob/1fa59756c2d7efa2dcd3e1130677ac2617bf6910/repo/packages/H/helloworld/0/marathon.json.mustache#L6

Maybe we should rename it to `dcos-http` and that would be consistent with the plugin functionality.

janisz:
@bamarni I'm not sure if that's the way to go. If this is a valid package then we should have a test for it.

br-lewis (approved):

drewvanstone (approved):

PR #2438, opened 2019-11-12
Make mysql local volume persist data

harryge00:
Currently, when `storage.persistence.enable` is enabled, we use the containerPath `var` to persist data. However, mysql does not write data to this directory, but instead, still writing to `/var/lib/data`.  So, after rebooting, the data will get lost.   The fix is based on https://docs.d2iq.com/mesosphere/dcos/1.12/storage/persistent-volume/#stateful-mysql-on-marathon

### DIFF
--- a/repo/packages/M/mysql/6/config.json
+++ b/repo/packages/M/mysql/6/config.json
@@ -1,0 +1,148 @@
+{
+	"properties": {
+		"service": {
+			"type": "object",
+			"description": "DC/OS service configuration properties",
+			"properties": {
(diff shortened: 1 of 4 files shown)


harryge00:
The marathon app created using the current package is like this:

image

We have a volume `var`, but it's not used because mysql still writes to `/var/lib/mysql`

kaiwalyajoshi (approved):

kaiwalyajoshi (approved):

PR #2452, opened 2019-11-28
Move from Docker to UCR for Confluent Replicator

wjglerum:
All other Confluent components use the UCR instead of Docker. Setting this to UCR allows us to monitor the memory usage for example too. Tested and verified on our test and production clusters.

### DIFF
--- a/repo/packages/C/confluent-replicator/200/marathon.json.mustache
+++ b/repo/packages/C/confluent-replicator/200/marathon.json.mustache
@@ -6,7 +6,7 @@
   "user": "{{replicator.user}}",
   "maintainer": "partner-support@confluent.io", 
   "container": {
-    "type": "DOCKER",
+    "type": "MESOS",
     "docker": {
       {{^replicator.virtual_network_enabled}}
       "network": "BRIDGE",


zmalik:
thanks @wjglerum!

rishabh96b:
Thanks for the PR @wjglerum ! I tested the `MESOS` container type and it was not working when  virtual network was not enabled because of the parameter `network: "BRIDGE"`. I will do some more changes and will cherry-pick your commit.

wjglerum:
Thanks for the feedback. We are running with MESOS and BRIDGE mode and that seems to work. What's going wrong?

rishabh96b:
With container changed to `MESOS` and keeping things as is in a non virtual network enabled scenario, I was getting
```
/container/docker/network should be empty and /container/docker/network should be empty
```
So I made the changes similar to that we have in other confluent packages [here](https://github.com/mesosphere/universe/pull/2453)
Can you check and let me know if these changes work for you also?

wjglerum:
Makes sense, I only edited the generated Marathon JSON directly, so I haven't used the templating from the package. Thanks for the info!

rishabh96b (commented):
Thanks again @wjglerum for the PR  and getting this to our notice. As already mentioned, I've raised a new PR #2453  and we can close this one

PR #2453, opened 2019-12-02
[DCOS-61537] Change container type to MESOS in replicator

rishabh96b:
This PR changes the container type `DOCKER` to `MESOS` as followed by all other confluent packages.

PR #2457, opened 2019-12-16
Revert "[DCOS-61537] Change container type to MESOS in replicator"

rishabh96b:
Reverts mesosphere/universe#2453 as the customer faced a broken replicator when running it in non VN mode.


zmalik (approved):
lets revert this!
I thought this was tested

rishabh96b:
Yes it was tested but I'll investigate why it broke on their side.

PR #2532, opened 2020-06-15
add a dcos-ui version to fix COPS-6139 on 1.13

pierrebeitz:
we want to provide a fix for COPS-6139 via the update service.

### DIFF
--- a/repo/packages/D/dcos-ui/6/package.json
+++ b/repo/packages/D/dcos-ui/6/package.json
@@ -1,0 +1,19 @@
+{
+  "packagingVersion": "4.0",
+  "name": "dcos-ui",
+  "version": "2.100.0",
+  "tags": ["dcos-ui", "dcos", "ui"],
+  "maintainer": "frontend-dev@mesosphere.io",
+  "description": "DC/OS UI",
+  "scm": "https://github.com/dcos/dcos-ui",
+  "website": "https://dcos.io/",
+  "framework": false,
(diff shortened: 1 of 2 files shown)


repo/packages/D/dcos-ui/6/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "packagingVersion": "4.0",
+  "name": "dcos-ui",
+  "version": "2.100.0",
+  "tags": ["dcos-ui", "dcos", "ui"],
+  "maintainer": "frontend-dev@mesosphere.io",
```
pierrebeitz, line 6: that address actually still works. not sure we have a better handle yet.

repo/packages/D/dcos-ui/6/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "packagingVersion": "4.0",
+  "name": "dcos-ui",
+  "version": "2.100.0",
```
pierrebeitz, line 4: we need a `2.x` version here that's bigger than `2.83.x` which is the version 1.13.7 shipped with because the update-service-ui will only show the update-button if the major version matches. so we introduce an artificial `2.100.0` here.

we'll soon be able to get rid of all versions `< 5.x`, as `5.x` will soon be capable of supporting all supported versions of dcos.

DominikDary (approved):

PR #2584, opened 2020-10-12
Deprecate Chronos

DominikDary:
Starting to deprecated Chronos. Jobs with dependencies has been recently added to Marathon and DC/OS 2.2.

@kaiwalyajoshi @timcharper @pierrebeitz please review.

/cc @vinodkone

### DIFF
--- a/repo/packages/C/chronos/7/package.json
+++ b/repo/packages/C/chronos/7/package.json
@@ -27,8 +27,8 @@
     }
   ],
   "preInstallNotes": "The use of chronos has been DEPRECATED! If you still choose to use it, we recommend a minimum of one node with at least 1 CPU and 2GB of RAM available for the Chronos Service.",
-  "postInstallNotes": "Chronos DCOS Service has been successfully installed!\n\n\tDocumentation: http://mesos.github.io/chronos\n\tIssues: https://github.com/mesos/chronos/issues",
-  "postUninstallNotes": "The Chronos DCOS Service has been uninstalled and will no longer run.\nPlease follow the instructions at https://docs.mesosphere.com/latest/deploying-services/uninstall/ to clean up any persisted state.",
+  "postInstallNotes": "[DEPRECATED] Chronos DCOS Service has been successfully installed!\n\n\tDocumentation: http://mesos.github.io/chronos\n\tIssues: https://github.com/mesos/chronos/issues",
+  "postUninstallNotes": "[DEPRECATED] The Chronos DCOS Service has been uninstalled and will no longer run.\nPlease follow the instructions at https://docs.mesosphere.com/latest/deploying-services/uninstall/ to clean up any persisted state.",
   "selected": false,
   "lastUpdated": 1513356170
 }
--- a/repo/packages/C/chronos/7/package.json
+++ b/repo/packages/C/chronos/7/package.json
@@ -4,7 +4,7 @@
   "version": "2.5.1",
   "scm": "https://github.com/mesos/chronos.git",
   "maintainer": "support@mesosphere.io",
-  "description": "A fault tolerant job scheduler for Mesos which handles dependencies and ISO8601 based schedules. [DEPRECATED]",
+  "description": "[DEPRECATED] A fault tolerant job scheduler for Mesos which handles dependencies and ISO8601 based schedules.",
   "framework": true,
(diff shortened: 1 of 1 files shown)


repo/packages/C/chronos/7/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "2.5.1",
   "scm": "https://github.com/mesos/chronos.git",
   "maintainer": "support@mesosphere.io",
-  "description": "A fault tolerant job scheduler for Mesos which handles dependencies and ISO8601 based schedules.",
+  "description": "A fault tolerant job scheduler for Mesos which handles dependencies and ISO8601 based schedules. [DEPRECATED]",
```
pierrebeitz: not sure if that's a pattern we already have in use. if so, that's fine with me. if not, i'd vote to make that more prominent. 

how about also adding it to the `preInstallNotes`?: 
```
  "description": "[DEPRECATED] A fault tolerant job scheduler for Mesos which handles dependencies and ISO8601 based schedules.",
  ...

  "preInstallNotes": "The use of chronos has been DEPRECATED! If you still choose to use it, we recommend a minimum of one node with at least 1 CPU and 2GB of RAM available.",

```

DominikDary: I had seen with other packages that some did it with the pattern I used, but I like your suggestions.

rishabh96b (commented):
Apart from @pierrebeitz 's suggestion, we should also add something like
```
  "postInstallNotes": "[Deprecated] Chromos DCOS Service has been successfully installed!",
  "postUninstallNotes": "[Deprecated] Chromos DCOS Service has been uninstalled.",
```

pierrebeitz (approved):

vinodkone (approved):
Would be nice if "deprecated" was a top level property of the json, but I guess we dont have that. cc @kaiwalyajoshi

pierrebeitz:
i strongly agree, @vinodkone. that was the first thing i checked. i haven't found it in universe and am 100% sure that we don't interpret something like that in the UI.

repo/packages/C/chronos/7/package.json
```diff
@@ -26,9 +26,9 @@
...
     }
   ],
-  "preInstallNotes": "We recommend a minimum of one node with at least 1 CPU and 2GB of RAM available for the Chronos Service.",
-  "postInstallNotes": "Chronos DCOS Service has been successfully installed!\n\n\tDocumentation: http://mesos.github.io/chronos\n\tIssues: https://github.com/mesos/chronos/issues",
-  "postUninstallNotes": "The Chronos DCOS Service has been uninstalled and will no longer run.\nPlease follow the instructions at https://docs.mesosphere.com/latest/deploying-services/uninstall/ to clean up any persisted state.",
+  "preInstallNotes": "The use of chronos has been DEPRECATED! If you still choose to use it, we recommend a minimum of one node with at least 1 CPU and 2GB of RAM available for the Chronos Service.",
+  "postInstallNotes": "[DEPRECATED] Chronos DCOS Service has been successfully installed!\n\n\tDocumentation: http://mesos.github.io/chronos\n\tIssues: https://github.com/mesos/chronos/issues",
+  "postUninstallNotes": "[DEPRECATED] The Chronos DCOS Service has been uninstalled and will no longer run.\nPlease follow the instructions at https://docs.mesosphere.com/latest/deploying-services/uninstall/ to clean up any persisted state.",
```
kaiwalyajoshi, line 31: Rather than just flagging this as `[DEPRECATED]` we should also directly point to the new Jobs & Dependencies feature in DC/OS 2.2.0 so users don't lose functionality.

kaiwalyajoshi (changes requested):

repo/packages/C/chronos/7/package.json
DominikDary: I think making that prominently visible is most important. 
Adding the note about DC/OS 2.2 makes sense, added it to the description.

kaiwalyajoshi (approved):

kaiwalyajoshi:
Building of `.dcos` file successful here: https://teamcity.mesosphere.io/viewLog.html?buildId=2967212&tab=buildResultsDiv&buildTypeId=DcOs_Enterprise_Components_PackageRegistry_UniversePackageBuilder_BuildCh

PR #2599, opened 2020-10-29
Release dcos-core-cli 2.2-patch.0

janisz:


janisz:
Nobody merged it over a week so I think it's fair to close

PR #2637, opened 2022-04-28
added mergebot workflow with test branch

shaeed-noordeen:


shaeed-noordeen:
secrets need to be added before we proceed.

kevinmiller-conley:
The new version of Mergebot is not currently supported in public repos, so I am closing this PR for now.

